PR #5, opened 2015-07-20
Remove dependency on `unstable-list-lib`

stamourv:
Depends on Racket PR #972, which moves most of its contents to core libraries.

Moves the contents of `unstable/bytes` to this package.


stamourv:
Merged.

PR #9, opened 2016-01-25
Add wrapper for radio input in radio-group

MarcKaufmann:
I replaced #:display by #:wrap (since the latter allows the former), although it is obviously not backwards-compatible.

### DIFF
--- a/web-server-doc/web-server/scribblings/formlets.scrbl
+++ b/web-server-doc/web-server/scribblings/formlets.scrbl
@@ -352,14 +352,16 @@ These @tech{formlet}s are the main combinators for form input.
 @defproc[(radio-group [l sequence?]
                       [#:attributes attrs (any/c . -> . (listof (list/c symbol? string?))) (λ (x) empty)]
                       [#:checked? checked? (any/c . -> . boolean?) (λ (x) #f)]
-                      [#:display display (any/c . -> . xexpr/c) (λ (x) x)])
+                      [#:display display (any/c . -> . xexpr/c) (λ (x) x)]
+					  [#:wrap wrap (any/c any/c . -> . xexpr/c) (λ (x y) (list x y))])
         (formlet/c any/c)]{
 
   This @tech{formlet} renders using a sequence of INPUT elements of
 RADIO type where each element gets its attributes from @racket[attrs]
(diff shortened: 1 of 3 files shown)


jeapostrophe:
You should add `display` back in for compatibility (also what you have right now has an error... the call to `display` that's still there is to the one that prints stuff out.) Can you also change the documentation (in web-server/web-server-doc/web-server/scribblings/formlets.scrbl) and the tests (web-server/web-server-test/tests/web-server/formlets-test.rkt) to use it?

MarcKaufmann:
Indeed the display was a blooper. I have now updated one test (checking that it inverts the order of the argument), but wasn't able to run it, since the (require ...) loaded the current form of radio-group, not my update. It looked fine in manual testing. I also updated the documentation, but no one has ever accused me of clear prose, so feedback is welcome.

I split each of the edits into a different commit (documentation, test, code), but if that's bad form, let me know, as I have only ever used Git for my own projects.

If I understand correctly, my new edits should all show up in the pull request.

jeapostrophe:
Awesome. Thank you Marc. It looks great.

samth:
This PR resulted in a failing test, which you can see here: http://drdr.racket-lang.org/33958/racket/share/pkgs/web-server-test/tests/web-server/run-all-tests.rkt

samth:
The failing test seems to be because the code change is to the `input-group` function, but the tests and docs changes are for the `radio-group` function.

MarcKaufmann:
Indeed, sorry about that. The changes in input-group are actually correct (or intentional, at least), but I didn't add the new #:wrap parameter to radio-group. I have done so now, shall I open a new pull request or is there a way to add to this one?

samth:
You can't add anything to this PR, since it's been merged. But I'm still confused. Did you accidentally change the wrong docs and tests? Or did something get left out of this PR?

MarcKaufmann:
I left something out of this PR, apparently. I didn't run the tests, as I couldn't figure out how to run tests with the code in my local repository.

And no, I changed the correct docs and tests (there are no docs for input-group).

I'll open a new pull request then, sorry for a noob mistake. I'll try and figure out how to run tests offline on my branch to avoid this type of mistake in the future.

PR #10, opened 2016-02-13
Add #:wrap to radio-group (was added in input-group)

MarcKaufmann:
Correcting the bug where only input-group had the #:wrap parameter, but not radio-group.

Again, didn't test it, as I need to figure out how to test my local branch.

PR #11, opened 2016-02-17
Update/correct radio-group contract.

MarcKaufmann:
This fixes the wrong contract that I built: `wrap` still returns a list of x-expressions, but the contract now reflects that. 

I tried to test the contract, but couldn't get it to fail on a (clearly wrong) contract, so this is again untested. 

I will try to find a way to run tests on my own branch, since me not doing so wasted quite a bit of time. Sorry about that.


samth:
I merged this, and used `xexpr-forest/c` as suggested by @jeapostrophe in #10.

PR #16, opened 2016-09-16
add code to parse content-disposition headers per RFC6266

jbclements:
Okay, that was a bit more unpleasant than expected. Lots of RFC-reading involved, and implementation of a teeny tiny parser combinator framework.

Also, what's our policy on unit tests? This pull request has a bunch of them.


jeapostrophe:
I'm very happy that you did it and that it works. I'm very anxious because of the big change and yet-another-parser library. There is a test suite, but I think it might be wise to solicit some tests from users of different browsers. I'm not trying to be a pain but I'm just trying to be conservative in this way.

jbclements:
I totally agree... I know plenty of Mozilians, let me see what I can do.

jeapostrophe:
We've been sitting on this for a long time. Any updates on the testing worry?

PR #17, opened 2016-12-14
Weaken the contract on response-output to permit any return value

lexi-lambda:
The contract on `response/output` demands that the provided function returns `void?`, which is needlessly restrictive, since it always ignores the result. Furthermore, this contract is actually even more pointless: it isn’t even enforced unless you use the `response/output` helper. The `response` struct is supposed to have the same contract on its `output`  field, but a typo in the contract uses `void` instead of `void?`, which permits any (single-valued) result.

This restriction is both annoying and actually confusing to people. It has come up on the mailing list before, and recently it came up again on the IRC channel. Furthermore, one of [Matt Might’s blog posts](http://matt.might.net/articles/low-level-web-in-racket/) actually depends on this permissive behavior of `response`, which is especially awkward given that it will fail if you change `response` to `response/output`.

This simply weakens the contract to what I think it probably should be, `(output-port? . -> . any)`, removing the possibility for confusion and being generally more permissive without any loss of safety.

### DIFF
--- a/web-server-doc/web-server/scribblings/http.scrbl
+++ b/web-server-doc/web-server/scribblings/http.scrbl
@@ -173,7 +173,7 @@ Here is an example typical of what you will find in many applications:
             [seconds number?]
             [mime (or/c false/c bytes?)]
             [headers (listof header?)]
-            [output (output-port? . -> . void)])]{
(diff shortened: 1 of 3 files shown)


jeapostrophe:
Thanks!

PR #19, opened 2017-03-07
Uses crypto-random-bytes for make-secret-salt/file

LiberalArtist:
in web-server/http/id-cookie

Also updates broken link to MIT Cookie Eaters


jeapostrophe:
Thank you!

PR #21, opened 2017-03-08
Use RFC 6265 cookies from net-cookies

### LINKED ISSUE #20: Cookies and RFC 6265
The `web-server/http/cookie` and `web-server/http/id-cookie` libraries (and, to a lesser extent, `web-server/http/cookie-parse`) are implemented using `net/cookie`, which carries a big scary warning about being deprecated in favor of `net/cookies` from the `net-cookies` package, which implements RFC 6265. 

I'm not sure what the web server should do about this. For my own purposes, I've more or less implemented the API of the web-server cookie libraries using `net/cookies`. I'm happy to clean up and share that code, but I'm not sure if it should be a package or a contribution to the web server and, if the later, what the backwards-compatibility considerations are. 

If the conclusion is that the web server should continue to provide the current libraries based on the deprecated `net/cookie`, it seems like the web server documentation should also note the limitations of `net/cookie` in appropriate places. It seems particularly relevant to `web-server/http/id-cookie`, as users probably want features like HttpOnly.

LiberalArtist:
Modifies web-server/http/cookie-parse, web-server/http/cookie,
and web-server/http/id-cookie to use net/cookies from the net-cookies
package, rather than the deprecated net/cookie.

resolves #20

### DIFF
--- a/web-server-doc/web-server/scribblings/http.scrbl
+++ b/web-server-doc/web-server/scribblings/http.scrbl
@@ -267,8 +267,10 @@ transmission that the server @bold{will not catch}.}
 @; ------------------------------------------------------------
 @section[#:tag "cookie"]{Placing Cookies}
 
-@(require (for-label net/cookie
-                     web-server/servlet
+@(require (for-label (except-in net/cookies/server
+                                make-cookie)
+                     net/cookies/common
+                      web-server/servlet
                      web-server/http/xexpr
                      web-server/http/redirect
                      web-server/http/request-structs
@@ -277,16 +279,25 @@ transmission that the server @bold{will not catch}.}
 
 @defmodule[web-server/http/cookie]{
  This module provides functions to create cookies and responses that set them.
-
- @defproc[(make-cookie [name cookie-name?] [value cookie-value?]
-                       [#:comment comment (or/c false/c string?) #f]
-                       [#:domain domain (or/c false/c valid-domain?) #f]
-                       [#:max-age max-age (or/c false/c exact-nonnegative-integer?) #f]
-                       [#:path path (or/c false/c string?) #f]
-                       [#:expires expires (or/c false/c string?) #f]
-                       [#:secure? secure? (or/c false/c boolean?) #f])
+ 
+ @defproc[(make-cookie [name cookie-name?]
+                       [value cookie-value?]
+                       [#:comment comment any/c #f]
+                       [#:domain domain (or/c domain-value? #f) #f]
+                       [#:max-age max-age (or/c (and/c integer? positive?) #f) #f]
+                       [#:path path (or/c path/extension-value? #f) #f]
+                       [#:expires expires (or/c date? string? #f) #f]
+                       [#:secure? secure? any/c #f]
+                       [#:http-only? http-only? any/c #f]
+                       [#:extension extension (or/c path/extension-value? #f) #f])
           cookie?]{
-  Constructs a cookie with the appropriate fields.
+   Constructs a cookie with the appropriate fields.
+
+   This is a wrapper around @racket[make-cookie] from @racketmodname[net/cookies/server]
+   for backwords compatability. The @racket[comment] argument is ignored.
+   If @racket[expires] is given as a string, it should match
+   @link["https://tools.ietf.org/html/rfc7231#section-7.1.1.2"]{RFC 7231, Section 7.1.1.2},
+   in which case it will be converted to a @racket[date?] value.
  }
 
  @defproc[(cookie->header [c cookie?]) header?]{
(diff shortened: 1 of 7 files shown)


web-server-doc/web-server/scribblings/http.scrbl
```diff
@@ -277,16 +279,25 @@ transmission that the server @bold{will not catch}.}
...
+                       [#:expires expires (or/c date? string? #f) #f]
+                       [#:secure? secure? any/c #f]
+                       [#:http-only? http-only? any/c #f]
+                       [#:extension extension (or/c path/extension-value? #f) #f])
           cookie?]{
-  Constructs a cookie with the appropriate fields.
+   Constructs a cookie with the appropriate fields.
+
+   This is a wrapper around @racket[make-cookie] from @racketmodname[net/cookies/server]
+   for backwords compatability. The @racket[comment] argument is ignored.
```
jeapostrophe: backwards

web-server-doc/web-server/scribblings/http.scrbl
```diff
@@ -277,16 +279,25 @@ transmission that the server @bold{will not catch}.}
...
+   If @racket[expires] is given as a string, it should match
+   @link["https://tools.ietf.org/html/rfc7231#section-7.1.1.2"]{RFC 7231, Section 7.1.1.2},
+   in which case it will be converted to a @racket[date?] value.
```
jeapostrophe: Does this mean it will fail if it does not? Does `net/cookies` enforce a contract?

jeapostrophe:
Thanks! Is it possible to add a few tests of the new arguments/features?

web-server-doc/web-server/scribblings/http.scrbl
LiberalArtist: `net/cookies/server` specifies `(or/c date? #f)`. The old implementation of `make-cookie` specifies `(or/c false/c string?)`, which is inherited from `cookie:add-expires` in `net/cookie`. So, `net/cookies/server` doesn't accept a string, the wrapper `make-cookie` will convert it if possible or throw an exception.

LiberalArtist:
Definitely — I haven't really looked at `web-server-test` yet, but I can later this afternoon.

LiberalArtist:
Working on the tests, I've found an inconsistency. Given this definition:
```
(define (make-reqc h)
   (make-request
    #"GET" (string->url "http://test.com/foo")
    (list (make-header #"Cookie" h)) (delay empty) #f
   "host" 80 "client"))
```
Evaluating `(cookie-header->alist (header-value (car (request-headers/raw (make-reqc #"$Version=\"1\"; name=\"value\"")))))` produces `'((#"$Version" . #"\"1\"") (#"name" . #"\"value\""))`. I use `cookie-header->alist` to implement the new `request-cookies`.

However, the old version of `request-cookies` eliminates the quotes: using the old definition, `(request-cookies (make-reqc #"$Version=\"1\"; name=\"value\""))` was expected to produce `'(#s(client-cookie "$Version" "1" #f #f) #s(client-cookie "name" "value" #f #f))`.

I'm not immediately sure what the source of the discrepency is or which behavior is correct.

LiberalArtist:
This test
```
(let ()
        (define in "hell\"w\"o") ;<--- this is not a cookie-value?
        (define out #"id=\"hell\\\"w\\\"o\"")
        (test-check "quotes (pr14194)" header-equal?
                    (cookie->header (make-cookie "id" in))
                    (make-header #"Set-Cookie" out))
        (test-equal? "quotes (pr14194)"
                     (reqc out)
                     (list (make-client-cookie "id" in #f #f))))
```
raises a contract violation because `cookie-value?` from `net/cookies/common` rejects values that contain double-quotes (or control characters, whitespace, commas, semicolons, or backslashes).

Should the test be eliminated? Or is `cookie-value?` wrong?

LiberalArtist:
Likewise, strings of the form `".foo"` do not satisfy `domain-value?`: they need to be of the form `"host.foo"` or just `"host"`.

LiberalArtist:
Ok, I have pushed some tests.

jeapostrophe:
The tests look good. Do you have anything else planned or should we commit now?

LiberalArtist:
I don't have anything else in mind, so I think it's ready to go.

jeapostrophe:
Thank you very much!

samth:
Unfortunately, this change causes significant problems for our continuous integration, because `net-cookies` depends on all of the Racket documentation. We therefore need to make some sort of change here. A few options:

1. Revert this change (bad).
2. Split the `net-cookies` package into `-lib`, `-test`, `-doc` variants.
3. Something else?

I prefer option 2, but I'm open to other suggestions.

LiberalArtist:
Splitting `net-cookies` seems reasonable to me, but I'm not 100% clear on how to do it. Is it just a matter of submitting a pull request to reorganize the git repository, or is there something the maintainer needs to do to update the package catalog?

samth:
Both are needed (for the latter, just pointing the name net-cookies at the
appropriate subdirectory, and creating the new packages)

LiberalArtist:
I've submitted a pull request to split `net/cookies`: [https://github.com/RenaissanceBug/racket-cookies/pull/6](https://github.com/RenaissanceBug/racket-cookies/pull/6)

jeapostrophe:
Can you prepare the pull-request for web-server to change its dep?

PR #22, opened 2017-03-15
Depend on net-cookies-doc and net-cookies-lib seperately

LiberalArtist:


jeapostrophe:
This must be applied at the same time as https://github.com/RenaissanceBug/racket-cookies/pull/6

jeapostrophe:
Also, @LiberalArtist a test case broke: http://drdr.racket-lang.org/39268/racket/share/pkgs/web-server-test/tests/web-server/http/xexpr.rkt

LiberalArtist:
Got it: `response/xexpr` was using the wrong `cookie?` in its contract.

samth:
One other question: should net-cookies now be part of main-distribution?

stamourv:
Since `net-cookies-lib` and `net-cookies-doc` will be by transitivity, I think that would make sense.
And `net-cookies-test` to `main-distribution-test`, too.

PR #24, opened 2017-04-19
Selective cache flushing

kimmyg:
This PR makes backwards-compatible changes to both `make-cached-url->servlet` and `cache-table-clear!` to allow caches to be flushed selectively.


jeapostrophe:
Excellent! Thanks Kimball!

PR #25, opened 2017-05-05
Adds a `#:shelf-life` argument to `valid-id-cookie?`

LiberalArtist:
Adds a `#:shelf-life` argument to `valid-id-cookie?` and `request-id-cookie`.

The current `#:timeout` argument does something counterintuitive:
it requires that the cookie have been created *before* a certain
moment in time, effectively specifying a *minimum* age.

The new `#:shelf-life` argument supports specifying a maximum age.


jeapostrophe:
Thanks!

samth:
This seems to have a broken test: http://drdr.racket-lang.org/40128/racket/share/pkgs/web-server-test/tests/web-server/http/cookies-test.rkt

LiberalArtist:
Thanks — I pushed https://github.com/racket/web-server/pull/26 to fix the test. (I'd overlooked that the test cases were using a cookie from a fixed time well over 500 seconds in the past.)

PR #28, opened 2017-09-17
Fix dispatcher/c and contract on next-dispatcher to use any

LiberalArtist:
The contracts had previously been written using `void` (n.b. *not* `void?`).
Similar to https://github.com/racket/web-server/pull/17


jeapostrophe:
Thank you!

PR #29, opened 2017-09-26
Fix formlet/c

LiberalArtist:
The old implementation of `formlet/c` could not actually handle multiple arguments. Expressions like `(formlet/c string? number?)` produced an error because `formlet/c` attempted to use `->` to define the processing function, but only `dynamic->*` can compute the number of results at runtime. The new implementation fixes this.

The new implementation is slightly more complicated because I wanted to avoid `dynamic->*` whenever possible, since it creates slower contracts than `->`. The obvious way is to make `formlet/c` a macro, but it was previously documented as being a function. I define `formlet/c` as a macro that expands to a function when it is used in a first-class context, and I have the function handle the one-argument case (i.e. all the uses of `formlet/c` that actually worked until now) specially.

I also revised the docs to clarify that `formlet*/c` can not be implemented in terms of `formlet/c`.

### DIFF
--- a/web-server-doc/web-server/scribblings/formlets.scrbl
+++ b/web-server-doc/web-server/scribblings/formlets.scrbl
@@ -202,7 +202,7 @@ types. Refer to @secref["input-formlets"] for example low-level formlets using t
  Equivalent to @racket[(listof xexpr/c)]
 }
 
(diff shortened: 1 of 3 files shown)


jeapostrophe:
Awesome! Good work!

PR #35, opened 2017-12-12
Add optional finalizer procedure to flushed servlet instances

kimmyg:
This PR adds an optional finalizer procedure both to the first result of `make-cached-url->servlet` and to `cache-table-clear!` in a backwards-compatible way.


jeapostrophe:
This is good and is backwards compatible. IMHO, the original behavior was an error and the default finalizer should be to shutdown the custodian. I'll merge this and you can let me know if you don't think that's right.

Issue #36, opened 2018-02-06
formlet/c leaves absolute paths in zo files


samth:
The implementation of `formlet/c`, in particular the use of `quote-this-module-path`, leaves that absolute module path in the resulting zo file. I think this can be fixed by using

```
(define quote-this-module-path (path->collects-relative-path (quote-module-path)))
```

but I'm not sure exactly what the intended goal is here, so I'm not sure.

PR #39, opened 2018-03-20
remove rpaths from formlets/compiled/input_rkt.zo

dbenoit17:
https://github.com/racket/web-server/issues/36


jeapostrophe:
The first part (removing the `.c`) should be a temporary thing until `expr/c` can be fixed. Can you submit a separate issue to the `syntax/parse` about this?

The second part (adding the collects-relative) can go in now. Can you separate them?

PR #44, opened 2018-11-09
fix: flush chunks to client after every iteration

Bogdanp:
This fixes an issue where, if the chunks of a chunked responses are too
short, they are never streamed to the client because they're buffered
within the OS socket.

A little more context here: https://racket.slack.com/archives/C06V96CKX/p1541782413390200


jeapostrophe:
Thank you!

Bogdanp:
@jeapostrophe My pleasure! I think there's another bug here, though this one is less important. When the client closes the connection (user closes the tab, for example), the chunking loop tries to write to the socket and it fails with `tcp-write: error writing`. I don't think it's a huge deal, but it is a little noisy and may surprise folks who don't know what's going on.

PR #45, opened 2018-11-10
Additional fixes and tests to streaming responses

Bogdanp:
I figured I'd fix the other bug I mentioned yesterday and also add some tests to prevent any future regressions. Let me know if you want me to change anything. I'm a Racket newbie so some of this might not be exactly idiomatic!


web-server-lib/web-server/http/response.rkt
```diff
@@ -109,26 +109,142 @@
...
-      (loop)))
-  (thread-wait to-chunker-t)
-  (fprintf to-client "0\r\n")
-  (fprintf to-client "\r\n")
-  (flush-output to-client))
+
+  ;; The client might go away while the response is being generated,
+  ;; in which case the output port will be closed so we have to
+  ;; gracefully back out when that happens.
+  (with-handlers ([exn:fail? (lambda (e)
```
Bogdanp: This ends up silences all `exn:fail`s, but I'm not sure what a better approach would be here since the error generated is something along the lines of `(exn:fail "display: port closed")`. Should I filter on the `exn-message` as well?

jeapostrophe:
This is a good direction. I am happy with the `with-handlers` you worry about in your comment. However, I prefer the tests to be in the `tests/web-server` directory of the `web-server-test` package at the top-level of this repo. Is that reasonable to you, to move it?

Bogdanp:
Yup! I'll move them to the test package tomorrow.

Bogdanp:
@jeapostrophe I've moved the tests into `web-server-test`.

jeapostrophe:
Excellent! Thanks!

Bogdanp:
@samth My bad. I ran `raco test web-server-test` rather than `racket web-server-test/tests/run-all-tests.rkt`, it looks like. I'll open a new PR this evening to fix the issue.

PR #47, opened 2018-11-15
fix test failures caused by previous PR

Bogdanp:
This fixes the test failures mentioned by @samth in #45.


jeapostrophe:
Thanks!

PR #51, opened 2019-01-17
checkpoint: fixup the defthings but introduce broken xrefs

bennn:
Work-in-progress, rewording the "Continue" tutorial to use `defthing`s properly (with a definition right below) or a different form instead.


jeapostrophe:
Do you want to continue with this or should we abandon it?

bennn:
abandon

PR #52, opened 2019-01-17
added @linebreak[] after signatures issue #50

spdegabrielle:
no changes to the text


jeapostrophe:
Thanks. We'll continue talking about other improvements on the other thread.

sorawee:
@jeapostrophe do you know why https://docs.racket-lang.org/continue/ is still not updated?

jeapostrophe:
That page always shows the last release and this patch came after merge day. The snapshot looks like what you expect: https://www.cs.utah.edu/plt/snapshots/current/doc/continue/index.html

PR #56, opened 2019-02-21
Add tests for generated URLs

jessealama:
Both `dispatch-rules` and `dispatch-rules+applies` yield functions that
generate URLs. There appear to be no tests to check what these URLs
are supposed to be like; here are a few.


jeapostrophe:
Thanks!

PR #57, opened 2019-02-27
redirect-to and redirect/get: support 307 redirects

LiberalArtist:
Previously, the `temporarily` status for `redirect-to`
meant the HTTP status `302 Found`, though this was
not guaranteed or documented.
For `302 Found` responses, user agents may use the same method as
the original request or may change the method from `POST` to `GET`.

This inconsistency is resolved by responding with `307 Temporary Redirect`
when the original method should always be used and
responding with `303 See Other` (which was already supported)
when the method should be changed from `POST` to `GET`,
e.g. for the Post-Redirect-Get pattern.

This commit changes the `temporarily` status for `redirect-to`
to mean `307 Temporary Redirect` and changes `redirect/get`
and `redirect/get/forget` to use the `see-other` status.

Also, the version of `redirect/get` from `#lang web-server`
is changed to accept a `#:headers` argument for consistency with
the stateful version.

This commit does not change the `permanently` status for `redirect-to`.
While `301 Moved Permanently` has the same issues as `302 Found`,
the replacement, `308 Permanent Redirect`, is not supported by
Internet Explorer on Windows 7 or 8.1
(though aparently IE on Windows 10 does support it:
see https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308).
The potential fallback behavior and workarounds
(see https://tools.ietf.org/html/rfc7538#section-4)
seem out-of-scope for this change.

### DIFF
--- a/web-server-lib/web-server/http/redirect.rkt
+++ b/web-server-lib/web-server/http/redirect.rkt
@@ -1,16 +1,22 @@
 #lang racket/base
 (require racket/contract
          (only-in racket/string non-empty-string?)
-         web-server/private/util
          web-server/http/response-structs
          web-server/http/request-structs)
 
-; redirection-status = (make-redirection-status nat bytes)
-(define-struct redirection-status (code message))
+; redirection-status = (redirection-status nat bytes)
+(struct redirection-status (code message))
 
-(define permanently (make-redirection-status 301 #"Moved Permanently"))
-(define temporarily (make-redirection-status 302 #"Moved Temporarily"))
-(define see-other (make-redirection-status 303 #"See Other"))
+(define permanently
+  ;; NOTE: 308 permanent redirect is not supported by
+  ;; Internet Explorer on Windows 7 or 8.1 as of 2019-02-26.
+  ;; (IE on Windows 10 does support it.)
+  ;; https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308
+  ;; https://tools.ietf.org/html/rfc7538#section-4
+  (redirection-status 301 #"Moved Permanently"))
+(define temporarily (redirection-status 302 #"Found"))
+(define temporarily/same-method (redirection-status 307 #"Temporary Redirect"))
+(define see-other (redirection-status 303 #"See Other"))
 
 ; : str [redirection-status] -> response
 (define (redirect-to 
@@ -19,7 +25,7 @@
          #:headers [headers (list)])
   (response (redirection-status-code perm/temp)
             (redirection-status-message perm/temp)
-            (current-seconds) #"text/html"
+            (current-seconds) #f
             (list* (make-header #"Location" (string->bytes/utf-8 uri))
                    headers)
             void))
@@ -31,4 +37,5 @@
  [redirection-status? (any/c . -> . boolean?)]
  [permanently redirection-status?]
  [temporarily redirection-status?]
+ [temporarily/same-method redirection-status?]
  [see-other redirection-status?])
--- a/web-server-doc/web-server/scribblings/http.scrbl
+++ b/web-server-doc/web-server/scribblings/http.scrbl
@@ -541,28 +541,94 @@ available (@racket[make-secret-salt/file]),
 
 @defmodule[web-server/http/redirect]{
 
-@defproc[(redirect-to [uri non-empty-string?]
-                      [perm/temp redirection-status? temporarily]
-                      [#:headers headers (listof header?) (list)])
-         response?]{
- Generates an HTTP response that redirects the browser to @racket[uri],
- while including the @racket[headers] in the response.
+@deftogether[                                   
+ (@defproc[(redirect-to [uri non-empty-string?]
+                        [status redirection-status? temporarily]
+                        [#:headers headers (listof header?) '()])
+           response?]
+   @defproc[(redirection-status? [v any/c]) boolean?]
+   @defthing[temporarily redirection-status?]
+   @defthing[temporarily/same-method redirection-status?]
+   @defthing[see-other redirection-status?]
+   @defthing[permanently redirection-status?])]{
+  The function @racket[redirect-to] 
+  generates an HTTP response that redirects the browser to @racket[uri],
+  while including the @racket[headers] in the response.
+  The @racket[status] argument is a @deftech{redirection status}
+  value, which determines the specific type of HTTP redirect to be used.
 
- Example:
- @racket[(redirect-to "http://www.add-three-numbers.com" permanently)]
-}
+  The default @tech{redirection status}, @racket[temporarily],
+  is preserved for backwards compatibility:
+  new code should usually use either @racket[temporarily/same-method]
+  or @racket[see-other], instead.
+  The @racket[temporarily] @tech{redirection status} corresponds to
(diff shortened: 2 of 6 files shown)


web-server-lib/web-server/http/redirect.rkt
```diff
@@ -8,8 +8,14 @@
...
-(define permanently (make-redirection-status 301 #"Moved Permanently"))
-(define temporarily (make-redirection-status 302 #"Moved Temporarily"))
+(define permanently
+  ;; NOTE: 308 permanent redirect is not supported by
+  ;; Internet Explorer on Windows 7 or 8.1 as of 2019-02-26.
+  ;; (IE on Windows 10 does support it.)
+  ;; https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308
+  ;; https://tools.ietf.org/html/rfc7538#section-4
+  (make-redirection-status 301 #"Moved Permanently"))
+(define temporarily (make-redirection-status 307 #"Temporary Redirect"))
```
Bogdanp: Isn't this technically a breaking change?

If I have a request handler that receives a `POST` request and I am using `redirect-to`, after this change, the browser will attempt to make a `POST` request to the page I am redirecting to, which may or may not support `POST` requests.

Pseudocode example of what I mean:


```racket
(define-values (dispatch _)
  (dispatch-rules
    [("login") #:method (or "get" "post") login-page]
    [("login-success") login-success-page]))

(define (login-page req)
  (cond
     [(and posted? valid-login?)
        (redirect-to "/login-success")]))

(define (login-sucess req)
  (response/xexpr '(h1 "Success")))
```

I believe this code would break after this change. 

IMO, this module should contain and expose the "new" redirection codes, but the default behaviour should continue being a 302.

LiberalArtist: I think there is a pragmatic question of whether existing programs might be relying on the behavior you illustrate.

On principle, my understanding is that the above code is relying on user-agent-defined behavior that is either not guaranteed (in HTTP 1.1, per [RFC 7231 §6.4.3](https://tools.ietf.org/html/rfc7231#section-6.4.3) or is explicitly called "erroneous" (in HTTP 1.0, per [RFC 1945 §9.3](https://tools.ietf.org/html/rfc1945#section-9.3). At the Racket level, `redirect-to` has not been documented as using any particular HTTP status code for the redirection.

On the pragmatic question, I always use `redirect/get` for the post-redirect-get pattern, so I assumed others did, too. If we think many people are using `redirect-to` with `temporarily` (explicitly or implicitly) and expecting it to work for post-redirect-get (or, at least, to work as well as it does now in most common user-agents), then changing the meaning of `temporarily` is probably wrong. In that case, I could see adding either two more statuses like `temporarily/same-method` or adding an optional keyword argument like `#:keep-method?` (but with a better name, so the default would apply to `301`, `302`, and `303`.

I don't see any reason for `redirect/get` to be using `302` instead of `303 See Other`, regardless.

jeapostrophe: @Bogdanp Can you check to see if it does break?

@LiberalArtist Can you do a quick search on github for uses of `redirect-to` that might be at risk? I checked in the Racket code I have access to and did not find any.

jeapostrophe:
In general, I'm weary of being too agitated towards following the letter of standards on the Web when the pragmatic behavior of browsers and the ecosystem is so different. @LiberalArtist are you experiencing a problem with the existing behavior or are you pursuing perfection?

web-server-lib/web-server/http/redirect.rkt
Bogdanp: FWIW, I do have (non-open source) code that does this exact thing. I realize `redirect/get` is better under many circumstances, but the UX of continuations is lacking when it comes to URLs that ought to be shareable between users, so I think this behavior would be missed for those types of situations.

@jeapostrophe it does break. Here's a minimal working example: https://gist.github.com/Bogdanp/4fce05956eba3838bd9a7aed3be48f83 . Change the 307 into a 302 and everything works as expected.

@LiberalArtist this is a fairly common pattern I've seen in languages other than Racket and  one I've used myself instead of relying on continuations in cases where the user must be able to easily share application URLs with others. All of the major browsers behave the same way with regards to HTTP 302 as far as I am aware. They are consistent with each other even though they may not necessarily be consistent with the spec.

> On the pragmatic question, I always use `redirect/get` for the post-redirect-get pattern, so I assumed others did, too. If we think many people are using `redirect-to` with `temporarily` (explicitly or implicitly) and expecting it to work for post-redirect-get (or, at least, to work as well as it does now in most common user-agents), then changing the meaning of `temporarily` is probably wrong. In that case, I could see adding either two more statuses like `temporarily/same-method` or adding an optional keyword argument like `#:keep-method?` (but with a better name, so the default would apply to `301`, `302`, and `303`.

My personal preference would be to add additional statuses and let the user choose which one is returned in the response. 

> I don't see any reason for `redirect/get` to be using `302` instead of `303 See Other`, regardless.

It's possible that clients that only support HTTP/1.0 won't know how to handle HTTP 303 responses. I don't mind either way, but is something to consider.

LiberalArtist:
I started down this road because I wanted to use redirects that don't change the method of the request. On reflection, though, I think I'm persuaded that it would be better to add a `temporarily/same-method` redirection status rather than change the meaning of `temporarily`. I will update this pull request to do that instead.

In terms of the documentation, if we think people are relying on the `redirection-status?` values to have the semantics of particular HTTP status codes, should we document what code each value means? I liked the idea of offering some abstraction over those details, but it sounds like the conclusion from this discussion is that dealing with the messiness of the web requires knowing about them. I think it would be odd to document the distinction between `temporarily` and `temporarily/same-method` without explaining `302 Found` vs. `307 Temporary Redirect`.

As far as `303 See Other` for `redirect/get`, I have always used 303 for post-redirect-get outside of Racket and have never had a problem. I was surprised when I saw that `redirect/get` didn't use `see-other` already. 303 redirects have been around since at least [RFC 2068](https://tools.ietf.org/html/rfc2068#section-10.3.4) in 1997. My impression is that it's one of the HTTP 1.1 features implemented even by some user-agents that don't implement all of HTTP 1.1 and announce themselves as HTTP 1.0—something similar to the `Host` header, which is really indispensable for navigating the modern web (virtual hosts break without it). [This article](https://blogs.msdn.microsoft.com/ieinternals/2011/08/19/http-methods-and-redirect-status-codes/) documents 303 working correctly from at least IE6, if not before, as well as the cross-browser inconsistency with `302 Found` for methods other than `GET` and (in practice) `POST`, even including `HEAD`: in addition to what the article lists, both [Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=56373) and [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=598304) subsequently changed their behavior so that they only rewrite `POST` to `GET`. I think `303 See Other` is the right choice for `redirect/get`.

jeapostrophe:
Okay, good feedback. I think the high-level take-away is that we shouldn't make breaking changes, but the documentation should clearly say what the response code for each status is. I don't think it is feasible to pretend that we are not putting out specific HTTP responses.

LiberalArtist:
I've pushed a new commit with the changes we discussed.

jeapostrophe:
Excellent, thank you!

PR #60, opened 2019-03-25
Fix some typos in the web-server docs

jessealama:
I spotted a handful of problems in the web-server docs:

* typos
* parallelism problems
* words that probably ought to be typed out (@tt) because they're referring to formal parameters of a function or command-line invocation


jeapostrophe:
Thanks!

PR #61, opened 2019-04-03
strengthen contracts on `response-code` and `response-seconds`

LiberalArtist:
Previously, both accepted `number?`, which was too permissive.

This commit adds `response-code/c`, which is defined as
`(integer-in 100 999)` per [RFC 2616 §6.1.1][1] ("The Status-Code
element is a 3-digit integer ...").

The contract for `response-seconds` is `real?`, which is the
range of `seconds->date` (which is used internally to generate the
`Last-Modified` header).

[1]: https://tools.ietf.org/html/rfc2616#section-6.1.1

### DIFF
--- a/web-server-doc/web-server/scribblings/formlets.scrbl
+++ b/web-server-doc/web-server/scribblings/formlets.scrbl
@@ -216,12 +216,18 @@ types. Refer to @secref["input-formlets"] for example low-level formlets using t
 
  (Actually, @racket[formlet/c] is a macro which avoids using @racket[dynamic->*]
  when the number of range contracts for the processing function is known at compile time.)
+
+ @history[#:changed "1.3"
+          "Fixed support for multiple return values."]
 }
 
 @defthing[formlet*/c contract?]{
   Similar to the contracts created by @racket[formlet/c], but uses @racket[any]
   to avoid checking the results (or even specifying the number of results)
   of the processing function.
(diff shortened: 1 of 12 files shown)


LiberalArtist:
This commit also bumps the version number for `web-server-lib` to 1.3, which probably should have been done before.

web-server-doc/web-server/scribblings/formlets.scrbl
```diff
@@ -505,6 +511,11 @@ a list of elements of the sequence.
 
 @defthing[input-int (formlet/c number?)]{
   Equivalent to @racket[(to-number input-string)].
+  Note that, despite the name, the result
+  @bold{is not guaranteed to be an integer.}
+  
+  @history[#:changed "1.3"
+           "Weakened result contract to allow any number."]
```
LiberalArtist: I do not understand the rationale behind <https://github.com/racket/web-server/commit/4cfb68e> (I couldn't find a PR 15447), but that is what it did.

LiberalArtist: I do not understand the rationale behind <https://github.com/racket/web-server/commit/4cfb68e> (I couldn't find a PR 15447), but that is what it did.

jeapostrophe:
Thank you, this is great

PR #62, opened 2019-04-05
Offer a variety of default standard status HTTP response messages

jessealama:
The various HTTP response-generating functions (`response/output`, `response/full`, and `response/xexpr`) currently require one to provide a status message, with `#"Okay"` serving as a default.

Supplying a status message in each call to `response/output` & friends is often inconvenient because, if one sticks to standard status codes, the status messages are usually a function of the code. In this pull request, I allow one to pass `#f` as the message. A message will be inferred from the code (which is required), with `#"OK"` serving as a fallback.

Existing code does not break. This change makes things more convenient going forward because one can omit one keyword argument in most cases. As before, it is always possible to supply a status message; if supplied, it will be used (the table of standard status messages will not be consulted).

The status codes and associated messages are taken from [RFC 7231](https://tools.ietf.org/html/rfc7231) and [RFC 7235](https://tools.ietf.org/html/rfc7235).

Additionally, there are some whitespace changes here (trimming up-to-end-of-line spaces).

### DIFF
--- a/web-server-doc/web-server/scribblings/http.scrbl
+++ b/web-server-doc/web-server/scribblings/http.scrbl
@@ -228,7 +228,7 @@ Examples:
                as the result (instead of demanding @racket[void?]).}]
 }
 
-@defproc[(response/full [code number?] [message bytes?] [seconds number?] [mime (or/c false/c bytes?)]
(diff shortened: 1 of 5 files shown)


jeapostrophe:
Very good. Thank you!

PR #67, opened 2019-07-30
Add new JSON convenience function à la response/xexpr

jessealama:
In this pull request we add a new convenience function for generating JSON-laden HTTP responses. It is to JSON what `response/xexpr` is to XML.


jeapostrophe:
Excellent. I've written something like this way too many times.

PR #70, opened 2019-08-27
server: gracefully handle "broken pipe" errors

Bogdanp:
Browsers (at least Chrome and Firefox) immediately close the
underlying socket whenever a user closes a tab, which leads to "broken
pipe" errors if response data is still in flight so this commit makes
it so that the dispatch server silently ignores those errors (as it
should).


Bogdanp:
> The error can happen on the read end if you kill the program sending a request (web server stress testing programs that you kill with Ctrl-C can do this).

I think I finally tracked down and fixed this in #76.

Issue #72, opened 2019-09-14
dispatch/servlet constructs an invalid manager


aymanosman:
The `instance-expiration-handler` of the default manager constructed by `dispatch/servlet` will returns an `xexpr?` instead of a `response?`.

```
Exception

The application raised an exception with the message:

make-threshold-LRU-manager: contract violation
  expected: can-be-response?
  given: '(html (head (title "Page Has Expired.")) (body (p "Sorry, this page has expired. Please go back.")))
  in: the range of
      a part of the or/c of
      the 1st argument of
      (->
       (or/c #f (-> request? can-be-response?))
       number?
       manager?)
  contract from: 
      <pkgs>/web-server-lib/web-server/managers/lru.rkt
  blaming: <pkgs>/web-server-lib/web-server/servlet-dispatch.rkt
   (assuming the contract is correct)
  at: <pkgs>/web-server-lib/web-server/managers/lru.rkt:16.2

Stack trace:

raise-blame-error16 at:
  line 347, column 0, in file /Users/ayman/Applications/Racket-v7.4/collects/racket/contract/private/blame.rkt
<unknown procedure> at:
  line 364, column 33, in file /Users/ayman/Applications/Racket-v7.4/collects/racket/contract/private/arrow-higher-order.rkt
select-handler/no-breaks at:
  line 163, column 2, in file /Users/ayman/Applications/Racket-v7.4/collects/racket/private/more-scheme.rkt
<unknown procedure> at:
  line 360, column 33, in file /Users/ayman/Applications/Racket-v7.4/collects/racket/contract/private/arrow-higher-order.rkt
<unknown procedure> at:
  line 63, column 2, in file /Users/ayman/m/packages/web-server/web-server-lib/web-server/dispatchers/dispatch-servlets.rkt
select-handler/no-breaks at:
  line 163, column 2, in file /Users/ayman/Applications/Racket-v7.4/collects/racket/private/more-scheme.rkt
<unknown procedure> at:
  line 144, column 8, in file /Users/ayman/m/packages/web-server/web-server-lib/web-server/private/dispatch-server-with-connect-unit.rkt
```

PR #73, opened 2019-09-14
Fix use of make-threshold-LRU-manager

aymanosman:
Fixes https://github.com/racket/web-server/issues/72


jeapostrophe:
Thank you!

Issue #74, opened 2019-09-14
Contract on make-id-cookie should accept byte-strings


LiberalArtist:
@soegaard reported on Slack that the contract on `make-id-cookie` currently requires the name and value arguments to be given as strings, so this program fails:
```racket
#lang racket

(require web-server/http/id-cookie
         racket/random)

(make-id-cookie #"aname" #"avalue"
                #:key (crypto-random-bytes 128))
```

All other cookie-constructing functions permissively accept byte-strings for names and values (converting them with `bytes->string/utf-8`), and it seems like `make-id-cookie` should, too—I probably just missed it after I figured out https://github.com/RenaissanceBug/racket-cookies/commit/c745ee51e9a4914d232380abe65332d91d88f7f8. I will try to make a PR soon.

LiberalArtist:
While working on a patch, I believe I found another bug. In this code: https://github.com/racket/web-server/blob/245047b620af75564ea14ac5d8ca9cae46d0ed04/web-server-lib/web-server/http/id-cookie.rkt#L83-L87
In https://github.com/racket/web-server/commit/a5135d5f1c96a21c55c704fa4bf56a4cc607307f, I believe @jeapostrophe should have changed `-3` to `-1` when he supplied `#""` to `base64-encode` for the newline instead of the default `#"\r\n"`.

A base64-encoded SHA-1 is always 28 bytes long, including exactly one `=` at the end for padding: since the length is fixed, we can drop the padding and return 27 bytes from `mac`. Right now, we are also dropping 2 significant digits of the SHA-1 hash.

I think this should be fixed and plan to include it in this patch. However, people's existing id cookies created with the current version of the library will not validate using a fixed version. I think the potential security implications outweigh the breakage, but I'm open to other opinions.

LiberalArtist:
If it seems necessary, one option would be to add a function `unverified-id-cookie?`. I think it would be like `valid-id-cookie?`, but `#:key` would be optional and default to `#f`. If the key is `#f`, it would just check the name and timeout and not try to verify the digest. beyond checking that it syntactically could have been generated by some version of this library (i.e. it is 25 or 27 base64 digits). Alternatively, if you supply a key, it would check the digest but allow it to be two digits short.

That would let programmers make their own decisions about how to handle their existing id cookies. But, personally, I don't feel a need for such a function.

jeapostrophe:
You should just change the `-3` to a `-1`. It was simply a mistake.

PR #75, opened 2019-09-15
web-server/http/id-cookie: accept names & values as bytes; fix HMAC

LiberalArtist:
Not ready to merge.
Needs docs, tests, and a look at `typed/web-server/http`.

### DIFF
--- a/web-server-doc/web-server/scribblings/http.scrbl
+++ b/web-server-doc/web-server/scribblings/http.scrbl
@@ -493,8 +493,8 @@ You can also generate random bytes using something like OpenSSL or @tt{/dev/rand
  @link["https://www.madboa.com/geek/openssl/#random-data"]{this FAQ} lists a few options.
 
  @defproc*[([(make-id-cookie
-              [name (and/c string? cookie-name?)]
-              [value (and/c string? cookie-value?)]
+              [name cookie-name?]
+              [value cookie-value?]
               [#:key secret-salt bytes?]
               [#:path path (or/c path/extension-value? #f) #f]
               [#:expires expires (or/c date? #f) #f]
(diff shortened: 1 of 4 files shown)


LiberalArtist:
Related to https://github.com/racket/web-server/issues/74

jeapostrophe:
Good work!

LiberalArtist:
I should be able to finish this up in time for 7.5, but it may not be until the weekend, depending on how some other things go. (Branch day is Monday.)

LiberalArtist:
Sorry this is so much later than I'd planned: I had some family obligations come up unexpectedly.

LiberalArtist:
I'm still thinking through what to do for `typed/web-server/http`:
- I'm leaning toward not allowing `Bytes` for name and value arguments. In the untyped API, bytes are accepted for a mixture of consistency, backward-compatibility (per <https://github.com/RenaissanceBug/racket-cookies/commit/c745ee5>), and "be liberal in what you accept." In the typed world, none of the cookie-related functions have ever accepted `Bytes`: changing that seems likely to cause more problems than it would solve, particularly in that it would make argument and result types different.
- I'm conflicted about strengthening the type of the `#:timeout` argument for `valid-id-cookie?` and `request-id-cookie` to `Real` instead of `Number`. (A non-real `#:timeout` has always caused a runtime error: this PR just corrects that in the documentation and, in one case, the contract.) On the one hand, we made essentially the same change to `response-seconds` in <https://github.com/racket/typed-racket/commit/acf7707>, and that seemed to go fine. On the other hand, it would cause some buggy programs that previously compiled to fail to typecheck: personally, if I'm using static types, it's because I want buggy programs to be rejected at compile-time, but in a strict sense it would be a breaking change. It may or may not be relevant that, usually, you don't actually want to be using the `#:timeout` argument (see <https://github.com/racket/web-server/commit/626ad02>).

jeapostrophe:
I don't have a strong opinion about the types. I follow your logic.

LiberalArtist:
Subsumed by <https://github.com/racket/web-server/pull/84>

PR #76, opened 2019-10-14
core: guard against RST packets during read

Bogdanp:
This prevents errors such as

    tcp-read: error reading
      system error: Connection reset by peer; errno=54
      errortrace...:
      context...:
       /Applications/Racket v7.4/collects/racket/port.rkt:1160:4: try-again

from appearing in application output after clients terminate their connections abruptly (without sending a FIN packet).

This code can be used to exercise the problem:

```racket
#lang racket

(require racket/tcp)

(for ([_ (in-range 100)])
  (define-values (in out)
    (tcp-connect "127.0.0.1" 8000))

  (parameterize ([current-output-port out])
    (displayln "GET / HTTP/1.1")
    (displayln "")
    (flush-output))

  (read-string 100 in)

  (tcp-abandon-port in)
  (tcp-abandon-port out))
```

### DIFF
--- a/web-server-lib/web-server/private/dispatch-server-with-connect-unit.rkt
+++ b/web-server-lib/web-server/private/dispatch-server-with-connect-unit.rkt
@@ -154,6 +154,16 @@
              (if (connection-close? conn)
                (kill-connection! conn)
                (connection-loop))])))))
+    (define liveness-bytes (make-bytes 1))
     (define (connection-loop)
-      (sync the-evt))
+      ;; If the socket at the other end is abandoned and we don't
+      ;; get sent a FIN packet, then we'll happily start to sync on
+      ;; the-evt. When that happens, `port-closed-evt' never fires and
+      ;; `peek-bytes-evt' encounters a RST packet inside its waiter
+      ;; thread (in `poll-or-spawn' in the runtime's `port.rkt') and
+      ;; that thread then dies with an exception that we can't guard
+      ;; against in here. Doing a non-blocking peek here avoids that
+      ;; whole scenario.
+      (unless (eof-object? (peek-bytes-avail! liveness-bytes 0 #f ip))
+        (sync the-evt)))
     (connection-loop)))
--- a/web-server-lib/web-server/private/dispatch-server-with-connect-unit.rkt
+++ b/web-server-lib/web-server/private/dispatch-server-with-connect-unit.rkt
@@ -135,7 +135,7 @@
     ;; consume a byte, etc.
     (define (connection-loop)
       (cond
-        [(eof-object? (peek-bytes 1 0 ip))
+        [(eof-object? (peek-byte/safe ip))
          (kill-connection! conn)]
 
         [else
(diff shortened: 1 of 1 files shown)


Bogdanp:
For reference, the thread in which the exception gets raised is here: https://github.com/racket/racket/blob/ca2eafbfc404207b4f8d5b1423dc1f7bb48afcef/racket/collects/racket/port.rkt#L1142-L1146 . Perhaps that code could also be updated to handle this case a bit better?

cc @mflatt

Bogdanp:
I realized I could simplify the whole connection loop by ditching all of the `evt` stuff. It looks like all the tests still pass and the original issue is still fixed. Was there a reason the loop was written that way (support for an older version of Racket?) or was it just entropy doing its thing?

mflatt:
I think there's confusion here about `port-closed-evt`. It's not about the other end of a connection-based port. It's about the port itself being closed concurrently in the same process with `close-input-port` or `close-output-port`. I imagine that the web server is set up so that the server may decide to terminate a connection by closing a port, but I don't really know.

Adding a `peek-bytes-avail!` before the `sync` causes `peek-bytes-avail!` to raise an exception if the port is closed with `close-input-port`, instead of letting the event created by `port-closed-evt` handle the close. So, assuming that the handling of a closed port (as opposed to a closed connection) is relevant, adding `peek-bytes-avail!` would not be the right thing.

Meanwhile, there really should be no difference to using `peek-bytes-avail!` and `sync` with `peek-bytes-evt` as far as the potential for `ECONNRESET` errors. For a given scenario, there is very likely a timing difference that can make one or the other trigger `ECONNRESET`, but neither one would consistently avoid `ECONNRESET` (unless I've missed something about how you're using them here).

Bogdanp:
> I think there's confusion here about port-closed-evt. It's not about the other end of a connection-based port. It's about the port itself being closed concurrently in the same process with close-input-port or close-output-port. I imagine that the web server is set up so that the server may decide to terminate a connection by closing a port, but I don't really know.

That's a good point I hadn't considered. I'll take another look at this tomorrow with that in mind just to be sure I didn't miss anything with the latter changes. Thanks!

> Meanwhile, there really should be no difference to using peek-bytes-avail! and sync with peek-bytes-evt as far as the potential for ECONNRESET errors. For a given scenario, there is very likely a timing difference that can make one or the other trigger ECONNRESET, but neither one would consistently avoid ECONNRESET (unless I've missed something about how you're using them here).

You're right that they are both susceptible to running into resets. The difference is that `peek-bytes-evt` swallows the exception because the exception gets raised in the thread it sets up to wait for bytes to become available ([this code](https://github.com/racket/racket/blob/ca2eafbfc404207b4f8d5b1423dc1f7bb48afcef/racket/collects/racket/port.rkt#L1142-L1146) which runs [this code](https://github.com/racket/racket/blob/ca2eafbfc404207b4f8d5b1423dc1f7bb48afcef/racket/collects/racket/port.rkt#L1160-L1209)). From the perspective of code that uses `peek-bytes-evt`, the event never becomes synchronizable (which might also mean that the connection leaks because of this, but I'm not 100% sure about that) and from the perspective of someone running the code, an error gets printed to standard out. Sorry I wasn't clear from the beginning!

mflatt:
Ah, I see. That does sound like a bug in `peek-bytes-evt`.

PR #79, opened 2019-11-08
ci: build against multiple Racket versions and variants

Bogdanp:
With this change, every push and pull request will be tested against both variants of Racket 7.4 as well as both variants of the latest snapshot.

### DIFF
--- a/web-server-test/tests/web-server/e2e/read-write/tests.rkt
+++ b/web-server-test/tests/web-server/e2e/read-write/tests.rkt
@@ -27,11 +27,15 @@
     "sending data too slowly"
     broken-pipe?
     (lambda _
-      (define-values (in out)
-        (tcp-connect "127.0.0.1" port))
+      ;; On Racket CS 7.4 the default plumber writes an error to
+      ;; standard out when it tries to close the socket. Creating a
+      ;; custom plumber seems to fix that problem.
+      (parameterize ([current-plumber (make-plumber)])
+        (define-values (in out)
+          (tcp-connect "127.0.0.1" port))
 
-      (parameterize ([current-output-port out])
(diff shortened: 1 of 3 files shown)


jeapostrophe:
I have no problem with this, but I don't know how the CI works, so I would just merge and see what happens. Maybe someone more knowledge would say that's dumb of me though?

web-server-test/tests/web-server/e2e/read-write/tests.rkt
```diff
@@ -27,11 +27,15 @@
     "sending data too slowly"
     broken-pipe?
     (lambda _
-      (define-values (in out)
-        (tcp-connect "127.0.0.1" port))
+      ;; On Racket CS 7.4 the default plumber writes an error to
+      ;; standard out when it tries to close the socket. Creating a
+      ;; custom plumber seems to fix that problem.
+      (parameterize ([current-plumber (make-plumber)])
```
samth: Is this error still there on RacketCS HEAD? Is it a bug?

Bogdanp:
@jeapostrophe you should be able to see a green checkmark next to my last commit. If you click on that, then you should be able to see the build status for each combination of racket version and variant. Clicking on the "details" link next to each of those will take you to the build output where you can see each command that was run as well as stdout and stderr.

web-server-test/tests/web-server/e2e/read-write/tests.rkt
Bogdanp: I think it might be. IIRC tests on the latest CS snapshot did not fail because of this, but I'm going to have to double check tomorrow.

samth:
I also looked at the change and it seems very reasonable.

jeapostrophe:
Thank you :)

PR #81, opened 2019-11-12
request: provide read-headers and parse-bindings

### LINKED ISSUE #80: `web-server/http/request` lost `read-headers`
It looks like `read-headers` disappeared from the provides of `web-server/http/request` sometime since the branch for v7.5. That breaks the "rfc6455" package, at least.

Bogdanp:
It looks like some packages (eg. rfc6455) were depending on these despite them not being documented.

Closes #80.

### DIFF
--- a/web-server-lib/web-server/http/request.rkt
+++ b/web-server-lib/web-server/http/request.rkt
@@ -26,6 +26,9 @@
 ;; Happy hacking!
 ;;
 
+(define positive-real/c
+  (and/c real? positive?))
+
 (define read-request/c
   (connection?
    listen-port-number?
@@ -34,6 +37,12 @@
    (values request? boolean?)))
 
 (provide/contract
+ [parse-bindings (-> bytes? (listof binding?))]
+ [read-headers (->* (input-port?)
+                    (positive-real/c
+                     positive-real/c)
+                    (listof header?))]
+
  [rename make-ext:read-request
          make-read-request
          (->* ()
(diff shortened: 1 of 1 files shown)


web-server-lib/web-server/http/request.rkt
```diff
@@ -34,6 +37,12 @@
    (values request? boolean?)))
 
 (provide/contract
+ [parse-bindings (-> bytes? (listof binding?))]
```
LiberalArtist: It might be worth adding a comment here so that others don't make the same reasonable assumption you did in the future.

samth: Or maybe even document these bindings.

web-server-lib/web-server/http/request.rkt
```diff
@@ -26,6 +26,9 @@
 ;; Happy hacking!
 ;;
 
+(define nonnegative-length/c
+  (or/c exact-nonnegative-integer? +inf.0))
```
LiberalArtist: Bikeshedding about names: if we use this for more things as we were discussing in <https://github.com/racket/web-server/pull/77>, it would apply to timeouts and counts (e.g. `#:max-request-files`) as well as "length"s per se. I don't have a great idea for a name, but maybe something like `nonnegative-limit/c`?

jeapostrophe:
I'm going to merge this quickly to get the package install workable and then we can do a documentation or argue about the names later.

PR #82, opened 2019-11-22
Make empty responses

jessealama:
In some HTTP API settings (e.g., JSON RPC) it is convenient to provide empty HTTP responses. This change adds a little convenience, `response/empty`, for such responses.

### DIFF
--- a/web-server-doc/web-server/scribblings/http.scrbl
+++ b/web-server-doc/web-server/scribblings/http.scrbl
@@ -381,6 +381,11 @@ Equivalent to
 Generates a response with an empty body. The usual @tt{Content-Type} header will be absent, unless passed in via @racket[headers]. Equivalent to
 @racketblock[(response code message seconds #f headers (λ (o) (write-bytes #"" o)))], with the understanding that if @racket[message] is missing (or @racket[#f]), it will be inferred from @racket[code] using the association between status codes and messages found in RFCs 7231 and 7235. See the documentation for @racket[response/full] for the table of built-in status codes.
 
+@history[
+  #:added "1.6"
+  @elem{First introduced.}]
+}
+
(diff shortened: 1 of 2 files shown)


jessealama:
I didn't bump the version number in info.rkt, but I'm happy to do that.

samth:
I think a version bump would be good, but more significantly a history annotation is needed.

jessealama:
@samth done.

I'm happy to update the Typed Racket web server after this changed has been approved & merged in.

LiberalArtist:
I would propose adding a `@history` not for version 1.6 but *not* actually incrementing the version in `info.rkt`, for a fairly specific reason. The changes in <https://github.com/racket/web-server/pull/77> referred to version 1.6 but also didn't increment the package version, which has the happy consequence of meaning no one can express a dependency on that API to the package system yet (and it isn't in the Racket 7.5 release), so we have the freedom to make the tweaks we've been discussing in that thread without breaking anyone.

I would suggest that we merge this straight-forward improvement now. I am working on a PR now that will:

- Incorporate the commit from <https://github.com/racket/web-server/pull/75>, which currently increments the version to 1.6;
- Make the change to a `safety-limits` struct I proposed in <https://github.com/racket/web-server/pull/77#issuecomment-552937123>; and
- Address some edge cases with `binding:file/port`, which I'm in the process of writing up now.

If we wait to increment the package version until then, we'll avoid any agita about compatibility during the intermediate steps.

jessealama:
I've bumped this to 1.6, but I'm happy to trash that commit @LiberalArtist.

LiberalArtist:
@jessealama I see our comments passed in the ether. That would be my vote, unless you or others disagree.

jessealama:
@LiberalArtist commit trashed

jeapostrophe:
Thanks @jessealama  I'll take this commit and hang tight for @LiberalArtist --- You guys are great!

PR #84, opened 2019-12-01
Fix compatability, id cookies, and binding:file; add safety-limits [1.6]

LiberalArtist:
I'm still cleaning up various things in the `safety-limits` commit—most notably, I haven't adjusted the tests—and finishing another commit for the `binding:file` edge cases, but I thought I'd open a PR now to solicit comments. See <https://github.com/racket/web-server/pull/77> and <https://github.com/racket/web-server/pull/75#issuecomment-552677557> for earlier discussion.

As I proposed in <https://github.com/racket/web-server/pull/82#issuecomment-557645643>, when we have consensus on the API changes, I plan to increment the `web-server-lib` version to `1.6` and open a corresponding PR in <https://github.com/racket/typed-racket> for `typed/racket/http`.


web-server-doc/web-server/scribblings/dispatch-server.scrbl
```diff
@@ -54,55 +68,274 @@ ports (e.g., to implement SSL) as used by the dispatch server.
...
+
+ The arguments to @racket[make-safety-limits] are used as follows:
+ @itemlist[
+ @item{The @racket[max-waiting] argument is passed to @racket[tcp-listen]
+   to specify the maximum number of client connections that can be waiting for acceptance.
+   When @racket[max-waiting] clients are waiting for acceptance, no new client connections can be made.
+   }
+  @item{The @racket[initial-connection-timeout] specifies the initial timeout,
+   in seconds, given to each connection.
+   }
```
LiberalArtist: @Bogdanp I think it would be useful to say more about how `initial-connection-timeout` interacts with the other timeouts, but I'm not 100% sure I've worked it all out myself.

LiberalArtist:
Here are some of the things that seem debatable:

## Response timeouts: `current-send-timeout` vs. connection manager changes

The `response-send-timeout` is used in `output-response` and friends from `web-server/http/response`, which are used in a number of different call sites. I liked that <https://github.com/racket/web-server/pull/77> required no source changes to uses of `output-response`-like functions, but I didn't communicating the value through a `current-send-timeout` parameter. Using a parameter required using `parameterize` around any context in which `output-response` might be called, which seemed error-prone—and indeed, <https://github.com/racket/web-server/pull/77> only parameterized `current-send-timeout` inside `serve` from `web-server/web-server`, so programs that used `serve/web-config@` or used the unit layer directly, for example, wouldn't set the parameter.

Instead, I extend `start-connection-manager` to accept an optional `safety-limits` argument and cooperate with `new-connection` to put the timeout inside an extended `connection` struct. This approach also seems extensible to future kinds of limits that apply to the connection. The downside is that the basic `connection` struct is documented, but creating connections without going through `new-connection` would circumvent the safety limits. It isn't entirely satisfying, but anyone who is manipulating connection instances can already do lots of bad stuff, so I've just documented the added restriction.

(The basic `connection` struct is also mutable, though the mutability isn't documented, and exploiting that seems like it already could cause problems painful to contemplate: for now I just ignore it, but a future PR might try to limit the mutability or at least not provide all of the mutators.)

## Configuration tables

How should configuration tables support `safety-limits`? In <https://github.com/racket/web-server/pull/77>, <https://github.com/Bogdanp/web-server/commit/cefe99f> extends configuration tables to support `response-timeout` and `response-send-timeout`, but not the other limits and timeouts. Here, I revert that change—some reasons are:

- Configuration tables remain consistent with `web-config^` (as opposed to the new `web-config*^`), rather than becoming a special case in which parts of `safety-limits` they support.
- In general, I've tried to make public APIs deal with whole `safety-limits` values, even when they only use a few pieces of them, in the interest of maintaining a single point of control and the possibility of adding new pieces in the future.
- Various mailing-list posts from @jeapostrophe suggest that the status of configuration tables is somewhere between "not recommended for new programs" and a sort of soft deprecation. [For example:](https://groups.google.com/d/msg/racket-users/igpDGJfdoTg/l31y67M6BwAJ), "You are right to use serve/servlet and friends rather than the
configuration-table stuff, which is just there for backwards
compatibility."  

## `serve/servlet` and `serve/launch/wait`

Before <https://github.com/racket/web-server/pull/77>, these supported only a `max-waiting` argument, whereas `serve` and friends from `web-server/web-server` also accepted an `initial-connection-timeout`. Right now, I don't add an `initial-connection-timeout` argument to functions that didn't already accept it, because adding a deprecated argument feels a bit silly. On the other hand, I can see a case for making everything consistent with `serve`.

## Miscellaneous

- Should any higher-level modules reexport `web-server/safety-limits`?
- Right now, I put the section documenting `safety-limits` in "Web Server: HTTP Server", §1 "Dispatching Servers", between "Dispatching Server Signatures" and "Dispatching Server Unit." I can think of various other places to put it: a new top-level section in "Web Server: HTTP Server"? In "Web Applications in Racket", since it's referenced there?

web-server-doc/web-server/scribblings/dispatch-server.scrbl
Bogdanp: I think the "timeout plan" at the top of `timer.rkt` is a fairly accurate description, although it should really mention which timeouts the timer gets reset to at every step. I think the `initial-connection-timeout` is completely superseded by the more fine-grained timeouts and is now only really used as the maximum time before the web server must start processing a request. Unless it's set to 0 or the server is under extreme load, it should never be reached.

jeapostrophe:
Re: Configuration Tables --- There are A LOT of features that these do not support and they have not been maintained in a very very long time, easily a decade. I don't think we should remove the command line tool, but it doesn't have feature parity now and couldn't feasibly have it.

web-server-doc/web-server/scribblings/dispatch-server.scrbl
LiberalArtist: Thanks for pointing me to this! From looking at the old behavior: https://github.com/racket/web-server/blob/ba300fde5f88a2d0883d37eaaba9ee140d425896/web-server-lib/web-server/private/timer.rkt#L107-L110

I think it would make more sense to remove `#:initial-connection-timeout` from `safety-limits`, pass deprecated `#:initial-connection-timeout` arguments as `#:request-read-timeout`, and set the initial timeout of the connection to the `#:request-read-timeout`. Does that seem right to you?

web-server-lib/web-server/http/request.rkt
```diff
@@ -589,7 +549,7 @@
 ;; This is kind of high just to be safe, but I don't think you ever see
 ;; more than a couple headers (Content-Disposition and Content-Type) per
 ;; field in practice.
-(define MAX-HEADERS/FIELD 20)
+(define MAX-HEADERS/FIELD 20) ;; ???
```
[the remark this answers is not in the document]
LiberalArtist: I'd meant to remove my comment—I hadn't initially realized that this applied to headers of a `multipart/form-data` part—but instead I'll note that in the comment lest anyone else be similarly confused.

I agree that this seems ample, especially given [RFC 7578 §4.8](https://tools.ietf.org/html/rfc7578#section-4.8). I'd leave `MAX-HEADERS/FIELD` as a constant.

web-server-lib/web-server/http/request.rkt
```diff
@@ -602,12 +562,14 @@
...
-                             #:max-fields [max-fields 100]
-                             #:max-field-length [max-field-length (* 8 1024)])
+(define (read-mime-multipart in boundary #:safety-limits [limits (make-safety-limits)])
+  (match-define (safety-limits #:max-request-files max-files
+                               #:max-request-file-length max-file-length
+                               #:request-file-memory-threshold max-file-memory-threshold
+                               #:max-request-headers max-fields
+                               #:max-request-header-length max-field-length)
+    limits)
+
```
LiberalArtist: I think this change was wrong: `#:max-request-headers` and `#:max-request-header-length` shouldn't control reading the body of `multipart/form-data` parts that end up as `binding:form` values. I will adjust `safety-limits` accordingly.

I think I may have been confused as with `MAX-HEADERS/FIELD` about "field," which obviously refers to form fields, but the phrase "MIME header fields" is also used in RFC 7478 discussing `multipart/form-data`.

web-server-lib/web-server/http/request.rkt
```diff
@@ -622,7 +584,7 @@
     (make-limited-input-port in n #f))
 
   (define (collect-part-headers)
-    (read-headers in MAX-HEADERS/FIELD (current-http-line-limit)))
+    (read-headers* in #:max-count MAX-HEADERS/FIELD #:max-length max-field-length))
```
LiberalArtist: Here I do think we should use `max-request-header-length`.

web-server-lib/web-server/http/request.rkt
```diff
@@ -712,7 +674,7 @@
       (reverse parts*)))
 
   (let skip-preamble ()
-    (define line (read-http-line/limited in))
+    (define line (read-http-line/limited in #:limit max-field-length))
```
LiberalArtist: I think this should also be `max-request-header-length`.

web-server-lib/web-server/http/request.rkt
```diff
@@ -651,8 +613,8 @@
         (define line
           (begin
             ;; Skip the CRLF and then read the boundary line.
-            (read-http-line/limited in)
-            (read-http-line/limited in)))
+            (read-http-line/limited in #:limit max-field-length)
+            (read-http-line/limited in #:limit max-field-length)))
```
LiberalArtist: @Bogdanp I'm less sure what the right limit is here. Previously it was `(current-http-line-limit)`, so `(* 8 1024)`.

web-server-doc/web-server/scribblings/dispatch-server.scrbl
Bogdanp: Yes, I think that sounds reasonable.

web-server-lib/web-server/http/request.rkt
Bogdanp: The call on line 616 should only ever read 2 bytes and the one after it should read up to an entire boundary line.  RFC2046 specifies that boundaries should be no longer than 70 characters so the total length of a boundary line can be up to 76 bytes long (two leading dashes + boundary + two final dashes (if final boundary) + CRLF). 

Since we know what the boundary is, we could do:

```racket
(read-http-line/limited in 2)  ;; CRLF
(read-http-line/limited in (+ (bytes-length end-boundary) 2))  ;; max boundary line length (including all the dashes) + CRLF
```

LiberalArtist:
I've pushed another batch of changes that are still broken. I'm aiming to wrap this up by the end of the week.

LiberalArtist:
It's looking like "the end of the week" was a bit overly optimistic, and I'm going to have limited connectivity for roughly the next week, but I should be able to finish not long after the new year.

web-server-lib/web-server/http/request.rkt
LiberalArtist: Does the `#:limit` to `read-http-line/limited` include the terminating CRLF? I think it does, in which case I think reading the CRLF should be `(read-http-line/limited in #:limit 0)`.

LiberalArtist: Yes, I tried it at the repl:
```
> (read-http-line/limited #:limit 2 (open-input-bytes #"--\r\n"))
#"--"
> (read-http-line/limited #:limit 0 (open-input-bytes #"\r\nabcdefg"))
#""
```

LiberalArtist:
I am having a hard time debugging one last test failure. [In the CI tests](https://github.com/racket/web-server/pull/84/checks?check_run_id=386427255#step:6:19), I am getting the failure `all-e2e-tests.rkt: raco test: non-zero exit: 1`. The most detail I've managed to get locally is:
```
Sapientia:web-server philip$ raco test web-server-test/tests/web-server/e2e/all-e2e-tests.rkt
raco test: (submod "web-server-test/tests/web-server/e2e/all-e2e-tests.rkt" test)
raco test: @(test-responsible '(jay))
all-e2e-tests.rkt: raco test: test raised an exception
```
I am not sure how to figure out more precisely what is going wrong.

With some tweaks to "all-e2e-tests.rkt" so I can comment out some cases, it looks like the `echo` and `tls` e2e tests are working fine, but all of the others have problems—but I still haven't gotten an actual error message.

There is one more substantive commit for the revisions to `binding:file`, and it is essentially done: I'm just holding it until I can get these tests to pass.

LiberalArtist:
I've squashed the history for this PR: the details are at <https://github.com/LiberalArtist/web-server/tree/wrap-up-1.6-history>.

Bogdanp:
> I am not sure how to figure out more precisely what is going wrong.

It's surprising that `raco test` doesn't report the exception. If I load the test submodule inside `racket-mode`, I get this exception:

```
; application: procedure does not expect an argument with given keyword
;   procedure: serve
;   given keyword: #:max-request-file-length
;   arguments...:
;    #:dispatch #<procedure:...patch-filter.rkt:12:0>
;    #:max-request-file-length 500
;    #:max-request-file-memory-threshold 250
;    #:max-request-files 2
;    #:port 9114
; Context:
;  /Users/bogdan/sandbox/web-server/web-server-test/tests/web-server/e2e/all-e2e-tests.rkt:41:6
;  /Applications/Racket v7.5/share/pkgs/rackunit-lib/rackunit/private/result.rkt:91:3
;  /Applications/Racket v7.5/share/pkgs/rackunit-lib/rackunit/private/test-suite.rkt:139:7
;  /Applications/Racket v7.5/share/pkgs/rackunit-lib/rackunit/private/test-suite.rkt:136:2
;  /Applications/Racket v7.5/share/pkgs/rackunit-lib/rackunit/private/test-suite.rkt:61:0 apply-test-suite
;  /Applications/Racket v7.5/share/pkgs/rackunit-lib/rackunit/text-ui.rkt:91:0 run-tests5
;  (submod "/Users/bogdan/sandbox/web-server/web-server-test/tests/web-server/e2e/all-e2e-tests.rkt" test):1:1 [running body]
; [Due to errors, REPL is just module language, requires, and stub definitions]
```

It looks like `e2e/file-upload/server.rkt` needs to be updated.

Bogdanp:
@LiberalArtist [this patch](https://gist.github.com/Bogdanp/223ff84f459265836c0d70707f9a5f59) gets the e2e tests passing.

LiberalArtist:
I think this is complete! (Finally!) Thanks @Bogdanp for walking me through how the new limits should work.

Bogdanp:
I've run the same attacks against this branch as I did against my own PR and I don't see any regressions, which is great! I also ran matchacha.ro's end-to-end test suite as well as played around with the website myself locally and everything is working fine. I'll do another pass over the changes and the new documentation in the morning, but this has a 👍 from me so far.

I'm not sure exactly what the branching process as part of a release is for this repo in particular, so I figured I'd ask: is there a chance that `master` will be released as part of `Racket 7.6` before this is merged, thereby making this (`safety-limits` in particular) a backwards-incompatible change with regards to that release? How do we make sure that doesn't happen? It may make sense to remove my changes from the 7.6 release if this can't get merged into the release branch "in time."

cc @samth

mflatt:
Whatever was in master as of January 7 is currently slated for the release. The manager for this repo, @jeapostrophe, can have anything he thinks appropriate replace that default until January 15. After January 15, any change requires a discussion among the release managers.

It sounds like Jay should decide whether to revert some commits to avoid backward-incompatibility (for now) or add new commits to get things into long-term shape.

jeapostrophe:
I think that we should add these commits for the long-term, unless someone sees any big problems by tomorrow, in which case, we should revert and continue working on this path.

PR #85, opened 2020-04-01
Move response/empty to http/request-structs.rkt.

jessealama:
The documentation incorrectly implies that `response/empty`
is in the `web-server/http/response-structs` module.  You
can see this if you look at

  https://docs.racket-lang.org/search/index.html?q=response%2Fempty

for 7.6. But this isn't true: if you use `response/empty` after `require`ing `web-server/http/response-structs`, you'll get an unbound identifier error. `response/empty` is actually provided from `web-server/http/empty`.

Rather than viewing this as a problem with the
documentation, the solution here is to incorporate
`response/empty`, which is currently defined in a little
module `empty.rkt`, into `response-structs.rkt`. It thus
sites alongside other utilities such as `response/full` and
`response/output`.

### DIFF
--- a/web-server-doc/web-server/scribblings/http.scrbl
+++ b/web-server-doc/web-server/scribblings/http.scrbl
@@ -430,18 +430,6 @@ Equivalent to
                as the result (instead of demanding @racket[void?]).}]
 }
 
-@defproc[(response/empty [#:code code number? 200]
-                         [#:message message (or/c false/c bytes?) #f]
-                         [#:cookies cookies (listof cookie?) '()]
-                         [#:seconds seconds number? (current-seconds)]
-                         [#:headers headers (listof header?) '()])
-         response?]{
-Generates a response with an empty body. The usual @tt{Content-Type} header will be absent, unless passed in via @racket[headers]. Equivalent to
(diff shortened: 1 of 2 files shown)


jessealama:
Pull request for the TR web-server is [here](https://github.com/racket/typed-racket/pull/922).

jessealama:
The solution here is just a proposal. I can imagine at least one alternative: provide `response/empty` from `web-server/http`. (A minor change to the docs would be needed.)

jessealama:
The build dies because TR is trying to use a file that gets deleted here. This change and the associated TR change depend on each other.

samth:
The re-provide solution seems simpler to me.

jeapostrophe:
I think the original intention was to provide it from `web-server/http` or correct the docs so that there's no backwards compatibility problems.

jessealama:
> I think the original intention was to provide it from `web-server/http` or correct the docs so that there's no backwards compatibility problems.

Sounds good. I've adjusted the docs and fixed the problem by (re)providing `response/empty` from `web-server/http`. I've squashed all the intermediate work into a single commit for the sake of cleanliness.

Backwards compatibility is preserved. `response/empty` never was really available in `web-server/http/response-structs`, despite what the docs implied, but it was available if you `require`d the undocumented `web-server/http/empty` module. That is preserved here.

jeapostrophe:
Thanks... this is a bit messy but I think this might be the right thing.

PR #87, opened 2020-04-10
dispatch-files: Gracefully handle non-GET & non-HEAD requests

### LINKED ISSUE #86: Dispatcher created with dispatch-files hangs on POST requests
If you make a dispatcher with `dispatch-files` and submit non-GET requests to it, you'll find that you don't get a response at all. The connection stays open, but the server doesn't send anything over the wire after emitting the headers. Consider this server, which is intended to serve up whatever is in the current working directory in which it is run:

````
#lang racket/base

; Dispatcher created with dispatch-files hangs (does not output response)
; on POST requests

(require
 web-server/http/xexpr
 web-server/web-server
 web-server/dispatchers/filesystem-map
 web-server/servlet-dispatch
 (prefix-in s: web-server/dispatchers/dispatch-sequencer)
 (prefix-in f: web-server/dispatchers/dispatch-files))

(define (fallback req)
  (response/xexpr '(html (body (p "Didn't work.")))))

(define dispatcher
  (s:make
   (f:make #:url->path (make-url->path (current-directory)))
   (dispatch/servlet fallback)))

(define stop-serving-cwd
  (serve #:dispatch dispatcher
         #:port 8080))

(with-handlers ([exn:break? (lambda (err) (stop-serving-cwd))])
  (sync/enable-break never-evt))
````

Actual: If you put a file, say, `foo.txt` in the directory in which you run this program, and do a GET request for it, you'll see the file. If you, say, POST to `/foo.txt`, you won't get any response at all; the server just hangs.

Expected: `fallback` gets called, and I see a "Didn't work." response.

The source of the behavior is in `output-file`, defined in `web-server/http/response.rkt`. There, in the definition of `output-file/boundary`, which is called by `output-file`, I see this comment:

````
; TODO: What if we want to output-file during a POST?
````

I propose tackling this TODO.

My suggestion is to keep the definition of `output-file` in `response.rkt` unchanged and focus on the static file dispatcher. I propose changing `dispatch-files.rkt` so that it accepts only three HTTP methods: GET, HEAD, and OPTIONS. Invoking the files dispatcher with any other method would trigger a call to `(next-dispatcher)`.

The proposed change would be largely backwards compatible. Method signatures wouldn't change. GET requests would continue to work as before. If one relied on HEAD and OPTIONS requests for static files *not* working, then this change would break that.

I'm happy to do the work here. I think the solution is straightforward to implement, but this page can serve as a place for discussing the issue.

Also worth thinking about:

* One could also consider dropping the HTTP method check in `output-file/boundary`. The function is essentially specialized on the `method` argument: it does something only when the method is `#"GET"` or some case-insensitive variation thereof.

jessealama:
closes #86

### DIFF
--- a/web-server-doc/web-server/scribblings/dispatchers.scrbl
+++ b/web-server-doc/web-server/scribblings/dispatchers.scrbl
@@ -346,10 +346,21 @@ Creates a denied procedure from an authorized procedure.
 
  This dispatcher supports HTTP Range GET requests and HEAD
  requests. If the request's method is neither HEAD nor GET,
- @racket[next-dispatcher] will be called.}
+ @racket[next-dispatcher] will be called.
 
-@history[#:changed "1.7"
-         @elem{Support for non-{GET,HEAD} requests.}]
+ If the path works out to something on disk (either as a
+ file, or, if the path refers to directory, one of the
+ specified @racket[indices] files in that directory), it
+ needs to be readable by the process that is running the web
+ server. Existing but unreadable files are handled as
+ non-existing files.}
+
+@history[
+  #:changed "1.7"
+  @elem{Support for non-{GET,HEAD} requests.}
+  #:changed "1.7"
+  @elem{Treat unreadable files as non-existing files.}
+]
 
 }
 
--- a/web-server-doc/web-server/scribblings/dispatchers.scrbl
+++ b/web-server-doc/web-server/scribblings/dispatchers.scrbl
@@ -344,7 +344,14 @@ Creates a denied procedure from an authorized procedure.
  Type of the path. The file is then
  streamed out the connection object.
 
- This dispatcher supports HTTP Range GET requests and HEAD requests.}}
+ This dispatcher supports HTTP Range GET requests and HEAD
+ requests. If the request's method is neither HEAD nor GET,
+ @racket[next-dispatcher] will be called.}
+
+@history[#:changed "1.7"
+         @elem{Support for non-{GET,HEAD} requests.}]
+
+}
 
 @; ------------------------------------------------------------
 @include-section["dispatch-servlets.scrbl"]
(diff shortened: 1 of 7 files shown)


jessealama:
Still missing: documentation updates and tests. I thought I'd share my current approach before digging into those other parts.

jeapostrophe:
The changes to `output-file` got reverted somehow, but what is still there is good and when I look at the older version, it looks like what I expected.

jessealama:
> The changes to `output-file` got reverted somehow, but what is still there is good and when I look at the older version, it looks like what I expected.

Right -- I believe I'm causing some confusion by mixing the discussion in the pull request (here) and the issue ([there](https://github.com/racket/web-server/issues/86)). What I wrote there is that I initially tried to work with the simple approach that just removes the method check in `output-files`, but found that that caused some tests to fail. Those failing test made it clear to me that just dropping the method check, in that place, would require some surgery that I think I'm not entirely prepared to undertake. Hence the new approach here.

jeapostrophe:
Ah, I understand now. So in that case, I'm good for this and will merge once there are docs/tests.

jessealama:
I opted for a simpler approach in which support for the OPTIONS method isn't baked in. The narrower solution is, essentially, to just call `next-dispatcher` is we're not dealing with HEAD or GET.

jessealama:
I'm sneaking in a few small changes with this pull request. I'm happy to treat these as separate topics, each with their own thread, but I thought that this is the right opportunity to deal with a few little things; I'd be curious to know what you think.

* `response/empty` probably ought to emit a 204 response (`No Content`) by default rather than 200. (It's still possible to supply your own status code, as it was before.)
* `response/empty` never has a body, so the response should probably have a `Content-Length` header with value `0`.
* `dispatch-files` throws an exception if you ask it to serve up an existing but unreadable file. Perhaps we ought to treat such files as non-existing (i.e., call `next-dispatcher`).

jeapostrophe:
I'm cool with all these. Thank you! And good work!

PR #89, opened 2020-04-18
New dispatcher constructor based on HTTP request method

jessealama:
The purpose of this change is to be able to filter requests by their method.

Motivation: Many servers don't handle arbitrary HTTP methods, but only a handful of some standard ones; and even then, a smaller sub-handful of methods are the only ones that are actually supported (e.g., GET and HEAD only).

One can use `dispatch-rules` and variants thereof for fine-grained dispatching based on the request method.  But, to be complete, one ought to provide a fallback in every case (e.g., a request-method-not-allowed servlet, where the method matches `(regexp ".*")`.). Adding a request-method-not-allowed rule for every URL in `dispatch-rules` can decrease code readability. With the kind of functionality introduced here, one can say, at the outset, that all requests have a known HTTP method (e.g, only GET and HEAD); nothing else will ever be considered.

Combining this dispatcher constructor with `dispatch-rules` and the sequencing dispatcher constructor makes for more readable code, in my view.

### DIFF
--- a/web-server-doc/web-server/scribblings/dispatchers.scrbl
+++ b/web-server-doc/web-server/scribblings/dispatchers.scrbl
@@ -166,6 +166,19 @@ URLs to paths on the filesystem.
 }}
 
 @; ------------------------------------------------------------
+@section[#:tag "dispatch-method"]{Filtering Requests by Method}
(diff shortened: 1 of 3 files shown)


jeapostrophe:
I like this a lot, thank you!

PR #90, opened 2020-04-20
test: deflake connection-manager-test

### LINKED ISSUE #88: Test occasionally fails
http://drdr.racket-lang.org/54239/cs/racket/share/pkgs/web-server-test/tests/web-server/run-all-tests.rkt

This particular test has been failing occasionally, but seemingly only on Racket CS.

```
--------------------
Connection Manager > Internal > Web Server > Extend timer
FAILURE
name:       check-true
location:   private/connection-manager-test.rkt:51:4
params:     '(#f)
--------------------
```

Bogdanp:
This potentially fixes #88.

There were two issues:

* the tests relied on timing to work correctly and
* the connection object could be garbage collected before the timers
ran because the tests never used the connections.

I've changed the tests so they now synchronize based on events and
laid out the code so that the connections should be retained at least
for as long as needed for their timers to fire.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        racket-version: ["7.5", "current"]
+        racket-version: ["7.5", "7.6", "current"]
         racket-variant: ["regular", "CS"]
     steps:
       - uses: actions/checkout@master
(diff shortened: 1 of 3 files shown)


Bogdanp:
@samth I did test with `parallel` under both variants. I performed the CS tests inside Docker w/ a `-j` of 5, though, because otherwise it kept getting oom-killed.

samth:
I tested this with 48 parallel versions, and it passed, so I'm going to merge.

PR #91, opened 2020-04-29
Allow symbol in dispatch-log's make to specify a logger

jessealama:
`dispatch-log`'s `make` expects a function for its log formatter. But the module also exports a utility, `log-format->format`, which takes a symbol and makes a log formatter. Presumably, the common way of invoking `dispatch-log`'s `make` is to call that helper with a symbol, thereby making a suitable log formatter function.

Why not spare a step there?

As a convenience, the change here allows us to specify a log formatter by giving a symbol, which is then passed to `log-format->format`. The symbol lookup (that is, the call to `log-format->format`) happens only once, when `make` is evaluated, so there's no interesting slowdown to be expected with this new convenience.

This change widens `dispatch-log`'s `make` contract, so the change should be backwards compatible.

Apart from that, there are some excess end-of-line whitespace trimmings here.

### DIFF
--- a/web-server-doc/web-server/scribblings/dispatchers.scrbl
+++ b/web-server-doc/web-server/scribblings/dispatchers.scrbl
@@ -254,15 +254,20 @@ a URL that refreshes the password file, servlet cache, etc.}
  @racket['apache-default] to @racket[apache-default-format].
 }
 
-@defproc[(make [#:format format format-req/c paren-format]
+@defproc[(make [#:format format (or/c log-format/c format-req/c) paren-format]
                [#:log-path log-path (or/c path-string? output-port?) "log"])
          dispatcher/c]{
  Logs requests to @racket[log-path], which can be either a filepath or an @racket[output-port?],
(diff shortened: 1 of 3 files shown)


jeapostrophe:
This is a fine idea. Can you add a test or two?

jessealama:
Sure, I’ll add a couple of tests.

jessealama:
I've added a test.

Unfortunately, it contains a flaky timeout. Without the timeout, the test fails (or, at least, I could never get it to pass; it may be that the `sleep` is necessary, though the length of the timeout is difficult to specify). I tested this on 7.6, CS variant. It seems to me that the channel used for processing requests and logging them needs a moment; if you drop the `sleep` in the test, you'll find that nothing has been logged, which is not the intended behavior.

jeapostrophe:
Thanks! I adjusted the test to use a pipe to avoid this problem.

jessealama:
Awesome, thanks for the help! I didn't realize I could use a pipe.

PR #92, opened 2020-05-04
Add response codes 206 and 308

jessealama:
Response code 206 is "Partial Content" and 308 is "Permanent
Redirect". According to the Mozilla trackers for these
codes,

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/206

and

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308

206 appears to be supported across the board. 308 is largely
supported, though some old browsers may do the wrong thing
when given these codes. (The docs already mention this
caveat, and it makes sense to keep the notes there
unchanged.)

Nonetheless, this change should be OK. Adding support for
these codes does not mean that the web server starts
producing unexpected 308 responses. Rather, the change
amounts to this: if the web server is asked to render a
response in which the code is 206 or 308, it generates a
suitable status message. It was, and remains, up to the
programmer to decide whether to use 308 at all, given that
it may not do the right thing for some older browsers.

### DIFF
--- a/web-server-doc/web-server/scribblings/http.scrbl
+++ b/web-server-doc/web-server/scribblings/http.scrbl
@@ -370,6 +370,7 @@ Examples:
    (list "203" "Non-Authoritative Information")
    (list "204" "No Content")
    (list "205" "Reset Content")
+   (list "206" "Partial Content")
 
    (list "300" "Multiple Choices")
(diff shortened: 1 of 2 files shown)


jeapostrophe:
Thank you!

PR #93, opened 2020-05-25
response: terminate failed responses early in responder

Bogdanp:
Norman Gray pointed out on the [mailing list][1] that responders don't gracefully handle the case where a response's output function throws an exception. Instead of truncating the response, they currently force the connection to hang until it times out or is closed by the client.

This PR "fixes" things so that responses are always terminated regardless of the state of the output function. I also liked @rmculpepper 's idea of intentionally breaking the response by writing a length w/o any follow-up content to make it clear that something went wrong to clients, but I'm not sure if any other web servers do that. I'll take a look at what other servers do tomorrow and report back.

After this change, this code:

```racket
#lang racket/base

(require web-server/http
         web-server/servlet-dispatch
         web-server/web-server)

(define stop
  (serve
   #:port 9111
   #:dispatch (dispatch/servlet
               (lambda (req)
                 (response/output
                  (lambda (out)
                    (displayln "hello" out)
                    (error 'failed)))))))

(with-handlers ([exn:break?
                 (lambda (e)
                   (stop))])
  (sync/enable-break never-evt))
```

Produces responses that look like this to the client:

```
$ telnet 127.1 9111
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
GET / HTTP/1.1

HTTP/1.1 200 OK
Date: Mon, 25 May 2020 20:35:44 GMT
Last-Modified: Mon, 25 May 2020 20:35:44 GMT
Server: Racket
Content-Type: text/html; charset=utf-8
Transfer-Encoding: chunked

6
hello

0

GET / HTTP/1.1

HTTP/1.1 200 OK
Date: Mon, 25 May 2020 20:35:51 GMT
Last-Modified: Mon, 25 May 2020 20:35:51 GMT
Server: Racket
Content-Type: text/html; charset=utf-8
Transfer-Encoding: chunked

6
hello

0
```

[1]: https://groups.google.com/d/msg/racket-users/NgNznbK_n3w/oG3ULVPDAQAJ

### DIFF
--- a/web-server-lib/web-server/dispatchers/dispatch-servlets.rkt
+++ b/web-server-lib/web-server/dispatchers/dispatch-servlets.rkt
@@ -96,5 +96,5 @@
                   (lambda ()
                     ((servlet-handler the-servlet) req))
                   servlet-prompt)))))))
-      
-      (output-response conn (any->response maybe-response)))))
+
+      (output-response/method conn (any->response maybe-response) (request-method req)))))
--- a/web-server-lib/web-server/http/response.rkt
+++ b/web-server-lib/web-server/http/response.rkt
@@ -141,9 +141,10 @@
     (thread (λ ()
               (with-handlers ([exn:fail?
                                (lambda (e)
-                                 ((error-display-handler) (exn-message e) e))])
-                ((response-output bresp) to-chunker))
-              (close-output-port to-chunker))))
+                                 (close-output-port to-chunker)
+                                 (raise e))])
+                ((response-output bresp) to-chunker)
+                (close-output-port to-chunker)))))
 
   ;; The client might go away while the response is being generated,
   ;; in which case the output port will be closed so we have to
--- a/web-server-lib/web-server/http/response.rkt
+++ b/web-server-lib/web-server/http/response.rkt
@@ -138,13 +138,23 @@
   (define-values (from-servlet to-chunker) (make-pipe))
   (define to-client (connection-o-port conn))
   (define to-chunker-t
-    (thread (λ ()
-              (with-handlers ([exn:fail?
-                               (lambda (e)
-                                 (close-output-port to-chunker)
-                                 (raise e))])
-                ((response-output bresp) to-chunker)
-                (close-output-port to-chunker)))))
+    (thread
+     (λ ()
+       ;; When errors occur, we immediately kill the connection to
+       ;; make it clear to the client that something went wrong.  By
+       ;; doing this, we ensure that the terminating chunk won't be
+       ;; sent and the client can't misinterpret the response as
+       ;; successful.  This is in line with how other web servers
+       ;; behave.
+       ;;
+       ;; See: https://github.com/racket/web-server/pull/93
+       (with-handlers ([exn:fail?
+                        (lambda (e)
+                          (kill-connection! conn)
+                          (close-output-port to-chunker)
+                          (raise e))])
+         ((response-output bresp) to-chunker)
+         (close-output-port to-chunker)))))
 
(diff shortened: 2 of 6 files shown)


rmculpepper:
Thanks for making this PR. One comment: If the purpose is to close the port, then I think this handler would be better:

    (lambda (e)
      (close-output-port to-chunker)
      (raise e))

That lets the exception be handled by the `uncaught-exception-handler`, whose default behavior is the same as your handler.

Bogdanp:
Good point. Thanks!

I took a look at a few web servers this morning:

* Python's [Gunicorn](https://gist.github.com/Bogdanp/767e70b24a5b3b823f8413ce81b96810) writes the first chunk of data, skips the terminating chunk and closes the connection.

* [uWSGI](https://gist.github.com/Bogdanp/258772bf42a56f49ad81c2e56a8c7e46) doesn't even attempt to use chunked encoding for the responses. If I modify the app so that it writes enough data to fill whatever buffering UWSGI does, then the result looks like [this](https://gist.github.com/Bogdanp/fe08a7a9d54f7ac25750b082317fba4c). Still no chunked encoding, but now the behavior is more similar to gunicorn in that some data is written and then the connection is closed.

* Ruby's [Puma](https://gist.github.com/Bogdanp/6c906f58428c6e1a99c4548025171ca0) writes the first chunk, then writes another status message and closes the connection.

* [nginx](https://gist.github.com/Bogdanp/9cbef2980bda0a3f01f4657bd2fbe95c) behaves just like gunicorn.

The Python app is [here](https://gist.github.com/Bogdanp/8be34b098dc7eda6786050086968ea1e), the Ruby app [here](https://gist.github.com/Bogdanp/bb81867b94e08f68cc93f4f09c3abf73) and the nginx configuration [here](https://gist.github.com/Bogdanp/122bb1e2c1352361be94aa6981093847).

So it seems like the common thing to do here is to abruptly terminate the connection when the error occurs, ensuring that the response can't be misinterpreted as valid because no terminating chunk is sent. 

I'll push a change to make the PR to make the server behave that way.

Bogdanp:
After my last change, the Racket app from my first comment now behaves like this:

```
$ telnet 127.1 9111
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
GET / HTTP/1.1

HTTP/1.1 200 OK
Date: Tue, 26 May 2020 09:12:30 GMT
Last-Modified: Tue, 26 May 2020 09:12:30 GMT
Server: Racket
Content-Type: text/html; charset=utf-8
Transfer-Encoding: chunked

6
hello

Connection closed by foreign host.
```

Bogdanp:
I pushed another commit to fix an issue I noticed with `HEAD` requests, namely that they currently send body data from servlets created using `dispatch/servlet`. Let me know if you'd like me to split that into a separate PR. Since `dispatch/servlet` executes continuations, it may make sense to change the managers to fail continuation lookups from `HEAD` requests, but I don't imagine this is a big problem at the moment.

jeapostrophe:
I'm happy with these changes and the tests. Do you see any other things to do or is it time to merge?

Bogdanp:
Yup, this is ready to merge. I'm not too worried about `HEAD` requests running continuations so I'll just open an issue for now so we don't lose track of the problem.

jeapostrophe:
Sounds great :)

jeapostrophe:
Sounds great :)

nxg:
For what it's worth, +1 from me!

I suspect there isn't a Correct Answer to this (and the only place to search for it might be on one of the W3C REST lists), but 'doing what other servers' do is probably the best solution over all.

PR #95, opened 2020-06-14
read-mime-multipart: fix content reading edge case

Bogdanp:
There is an edge case in the current implementation of `read-mime-multipart` that causes it to miss boundaries when the data is split in the middle of a boundary and the next boundary is in the buffer after the split. I've updated the tests to randomly chunk the fixtures, which made the error apparent, and fixed the implementation.


jeapostrophe:
Thank you for your hard work!

PR #103, opened 2021-07-12
add filesystem:errno to expected exceptions (for unix domain sockets)

### LINKED ISSUE #102: Unhandled exceptions when using unix-domain-socket tcp unit
When using the unix-domain-socket tcp unit, broken pipe (errno 32) exceptions are not being caught and handled because they have type `exn:fail:filesystem:errno` , and the dispatcher is only handling `exn:fail:network:errno`

This is only an annoyance (exception spam in the console)

djholtby:
Adds filesystem errno exceptions to the list of expected (harmless) exceptions, so that broken-pipe is correctly handled when coming from a unix domain socket.

Fixes #102


jeapostrophe:
Thanks!

PR #104, opened 2021-07-13
ci: update racket versions and pin setup-racket@v1.5

Bogdanp:
I noticed in #103 that CI was failing due to some changes to the installer URI scheme so this fixes that problem.


jeapostrophe:
Thanks!

PR #106, opened 2021-07-14
remove run-server

samth:
Not yet sure if this is a good idea.

### DIFF
--- a/web-server-lib/web-server/private/dispatch-server-with-connect-unit.rkt
+++ b/web-server-lib/web-server/private/dispatch-server-with-connect-unit.rkt
@@ -1,7 +1,6 @@
 #lang racket/unit
 
-(require mzlib/thread
-         net/tcp-sig
+(require net/tcp-sig
          racket/async-channel
          racket/match
          "../safety-limits.rkt"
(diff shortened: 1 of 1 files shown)


jeapostrophe:
We just don't like mzlib? Is this just a copy & paste of `run-server`?

samth:
Copy-paste and specialized. 

This removed the dependency on mzlib entirely for the web-server-lib package. I think the other possibility is moving run-server somewhere more sensible.

LiberalArtist:
I remember looking at `run-server` when I was working on https://github.com/racket/web-server/pull/84 and being surprised, first that such a core part of the web server was implemented in a deprecated library from the `compatibility-lib` package, then that `run-server` was in a deprecated package, since it seemed to handle some tricky details (much of what [“More: Systems Programming with Racket”](https://docs.racket-lang.org/more/index.html) covers, IIRC) that could be nice to reuse, e.g. for non-TCP servers.

Putting `run-server`, or the most useful parts of it, in some non-deprecated package is appealing, especially if it were modernized a bit, e.g. with keyword arguments.

On the other hand, it's hard to know when to stop generalizing and modernizing, especially without having a second use concretely at hand. For example, having the kill timeout be hard-coded as twice the break timeout seems unfortunate for a general function.

If it seems better to copy–paste for now, I'd add a comment here and maybe even in `mzlib/thread` in the hope of prompting someone to tackle the refactoring in the future.

Bogdanp:
I think this can be closed now that #121 is merged.

PR #108, opened 2021-10-14
Updated to MIT link

ghost:
Note the MIT link for an HTML page is seemingly gone and replaced with a PDF version.   I wasn't sure if there was a requirement to only link to HTML pages but the only other options for this article were clones of the page on various sites that I didn't recognize the TLD for so I opted to fix with a link from the original source even though it's a PDF.


jeapostrophe:
Thanks!

PR #109, opened 2021-12-11
Make it possible to support a caching policy for the files dispatcher

jessealama:
The file dispatcher is a servlet that serves up files on disk as opposed to dynamically generated content. Such content is allowed to change, but it typically changes at a comparatively (much) slower rate than the dynamically generated parts of a web site. Servlets may be changed at will, but files sitting in a directory often remain unchanged from deployment to deployment. Such content is normally said to be static.

For static content, it is common (if not advisable) to have a caching policy. The [standard](https://www.ietf.org/rfc/rfc2616.html#section-14.9) `Cache-Control` HTTP response header is the means for expressing such a policy. A [small zoo](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) of standard values are available for that header. The purpose of the change here is to make it possible for file dispatchers to support these values.

(Having a caching policy even for the dynamically generated part of a web app is meaningful, but that topic of scope for this discussion, which is limited to serving files on disk. Other content besides files on disk might also be said to be "static", but again, we are talking only about files on disk here. Also, files on disk may, in some web applications, be far from "static", but that's OK because the status quo is that the file dispatcher currently does not add any cache-related headers to its responses, and with this change, that remains possible. Indeed, that remains the default.)

This change is backwards compatible because all the new keyword arguments to `dispatch-files`'s `make` function are optional, and have a default value of `#f`, which has the effect that a dispatcher will not emit the `Cache-Control` header.  Only when a non-`#f` value is passed for a cache-related keyword argument will a `Cache-Control` header be emitted. ~~`output-file` in `web-server/http/response` is changed in a backwards-incompatible way (a new mandatory argument is added), but that function is not `provide`d outside of `web-server`, so the change shouldn't be detectable. Indeed, the only use of `output-files` I could find is, naturally enough, in the file dispatcher.~~


jessealama:
There are also a number of non-essential whitespace changes here (usually excess end-of-line whitespace) that I'd like to commit. My editor automatically trims such whitespace upon saving files, which is why they show up in the diffs. I didn't intentionally seek out whitespace issues.

jeapostrophe (approved):

jeapostrophe:
This looks good, do you expect any more changes?

jessealama:
No more changes expected; I’d say this is good to go.

PR #110, opened 2021-12-19
New wrapping dispatcher for transforming requests and responses

jessealama:
A common pattern in web development is adding layers of request and response wrapping. That is, a web app often doesn't handle the original incoming request that hit the web server. Rather, it handles an enriched or otherwise transformed request. Similarly, the final response that ultimately gets serialized over the network may not be the response that the app emitted (after any initial request request transformations). As with requests, the final response may be the result of a chain of transformations.

As an example of all this, consider the task of injecting headers. Headers may be added (or changed, or dropped) from a request to store computed values that are needed further down the line. Authentication, for example, or processing cookies fits this pattern.

Another class of examples concerns bodies. Request and response bodies may be similarly intercepted. A response body, for example, might be parsed so that and obscenities or sensitive information can be filtered out. Cookies may be set.

Here we define a new kind of dispatcher that encapsulates this wrapping pattern. Given a servlet (that is, a function from `request?` to `response?`), a request transformer (`request? . -> . request?`) and a response transformer (`response? . -> . response?`), a `dispatcher/c` is created that passes the servlet the transformed request, and transforms the response returned by the servlet.

Note that in the documentation there's a reference to version 1.9. That's the same version that introduced the ability to set caching-related headers in the files dispatcher. It seems excessive to me to bump the version again, to 1.10, just for this change. In my view these two changes could be bundled together for the 1.9 release.

### DIFF
--- a/web-server-doc/web-server/scribblings/dispatchers.scrbl
+++ b/web-server-doc/web-server/scribblings/dispatchers.scrbl
@@ -511,3 +511,45 @@ Consider this example:
 
 (do-not-return)
 ]
+
+@; ------------------------------------------------------------
+@section[#:tag "wrap"]{Wrapping Requests & Responses}
+@a-dispatcher[web-server/dispatchers/dispatch-wrap
+              @elem{provides a general-purpose wrapping dispatcher that allows one to intercept an incoming request as well as the response returned by other servlets}]{
+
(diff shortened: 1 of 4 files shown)


jeapostrophe:
Thanks!

PR #112, opened 2021-12-20
Fix contract for `response/output`

### LINKED ISSUE #111: Error in contract for `response/output`
The contract for `response/output` is incorrect: the `#:message` keyword argument should be `(or/c bytes? #f)` rather than just `bytes?` as it currently is.

jessealama:
The contract was overly strict. The intention all along was
to allow `#f` as the `#:message` argument. That's what's in
the docs, and indeed, we use `#f` in some tests.

closes #111


jeapostrophe:
Thank you!

PR #117, opened 2022-06-03
Add #:path->headers to dispatch-files

sorawee:
We might want to set the response header based on path.
For example, a server might want to serve a gzip file (.gz) with
`Content-Encoding: gzip`, so that browsers automatically
unzip the file on the client side.


jeapostrophe (approved):

jeapostrophe:
Thank you!

PR #119, opened 2022-06-11
Add current-header-handler and dispatch-logresp

### LINKED ISSUE #54: Add response logging
Suppose I have a web application:

```racket
#lang racket
(require web-server/servlet-env)

; Some web app code here ...

;; Start the web server.
(serve/servlet request-handler
                       #:log-file "/dev/stdout")
```

When I start the web server and access the web app through the browser, the logs appear like this:

```
127.0.0.1 - - [25/Dec/2018:12:34:56 +0000] "GET /servlets/standalone.rkt HTTP/1.1" - -
127.0.0.1 - - [25/Dec/2018:12:34:56 +0000] "GET /favicon.ico HTTP/1.1" - -
```

The HTTP response code is missing from the logs, making it difficult to know when any of the web pages result in `500`, `403`, `404`, etc.

Would it be possible to implement request + response logging? Request + response logs are more useful in tracking past errors. For example:

```
127.0.0.1 - - [25/Dec/2018:12:34:56 +0000] "GET /servlets/standalone.rkt HTTP/1.1"  401 152 "-" "Mozilla/5.0 (X11; Linux x86_64; rv:60.0) Gecko/20100101 Firefox/60.0"
127.0.0.1 - - [25/Dec/2018:12:34:56 +0000] "GET /favicon.ico HTTP/1.1" 404 152 "-" "Mozilla/5.0 (X11; Linux x86_64; rv:60.0) Gecko/20100101 Firefox/60.0"
```

(Originally asked on Stack Overflow: https://stackoverflow.com/questions/53911162/)

sorawee:
Add a new parameter `current-header-handler` which can be used
to transform the "header" (including code, message -- everything but the
body) right before it will be written to the output port.

Use this new parameter to implement `dispatch-logresp`, a new dispatcher
that consumes another dispatcher, and returns a dispatcher that acts
like the input dispatcher, except that if the input dispatcher is
dispatched, it will log request and response, just like how the
Apache server does.

Still need tests and documentation.

Partially fixes #54

### DIFF
--- a/web-server-lib/web-server/dispatchers/dispatch-logresp.rkt
+++ b/web-server-lib/web-server/dispatchers/dispatch-logresp.rkt
@@ -1,17 +1,15 @@
 #lang racket/base
-(require net/url
-         (prefix-in srfi-date: srfi/19)
-         racket/date
-         racket/async-channel
-         racket/match
-         racket/contract)
-(require web-server/dispatchers/dispatch
+
+(require racket/contract
+         web-server/dispatchers/dispatch
          web-server/http
-         web-server/http/response)
+         web-server/http/response
+         "private/log.rkt"
+         (submod web-server/dispatchers/dispatch-log private))
+
 (define format-reqresp/c
   (or/c (-> request? string?)
         (-> request? response? string?)))
-(define log-format/c (symbols 'parenthesized-default 'extended 'apache-default))
 
 (provide/contract
  [format-reqresp/c contract?]
@@ -26,11 +24,13 @@
              #:log-path (or/c path-string? output-port?))
             dispatcher/c)])
 
-(define ((log-header-handler log-message req original-handler) resp)
-  (log-message req resp)
-  (original-handler resp))
-
 (define interface-version 'v1)
+
+(define ((log-header-handler log-message req original-handler) resp)
+  (define new-resp (original-handler resp))
+  (log-message req new-resp)
+  new-resp)
+
 (define (make #:format [format paren-format]
               #:log-path [log-path "log"]
               dispatcher)
@@ -38,7 +38,13 @@
     (if (symbol? format)
         (log-format->format format)
         format))
-  (define log-message (make-log-message log-path final-format))
+  (define log-message (make-log-message
+                       log-path
+                       (λ (req resp)
+                         (cond
+                           [(procedure-arity-includes? final-format 2)
+                            (final-format req resp)]
+                           [else (final-format req)]))))
   (lambda (conn req)
     (with-handlers ([exn:dispatcher? (lambda (e) (next-dispatcher))])
       (parameterize ([current-header-handler (log-header-handler log-message req (current-header-handler))])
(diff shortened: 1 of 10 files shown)


sorawee:
@jeapostrophe I don't know if this is a "clean" enough for you. I handle the code at a lower level than you suggested, because I also want it to be able to log response code from `dispatch-files`, etc.

Some concerns:
- Is there a good way to test how performance is affected by the change? (perhaps @Bogdanp know?)
- The name `header` might be confusing, because it could mean "header?" or "the part of response? that does not include body"
- Is it a good idea to reuse the `response` struct (with the `output` field being emptied)?

LiberalArtist:
I've sometimes wanted (and have sometimes implemented) ways to interpose on response handlers.

I haven't reviewed the code in detail—maybe you've already thought about this!—but one thing I wanted to flag is the use of a parameter for `current-header-handler`: it's important to think through the details of how parameters interact with continuations and whether that gives you the semantics you want. I was reminded [recently on Discourse](https://racket.discourse.group/t/running-a-procedure-from-delay-does-not-capture-modified-parameters/936/10) of an old, bad bug I wrote based on a faulty understanding of that interaction, and I'm pretty sure the buggy implementation—everyone who logged in to the site ended up in the account of the first person to log in after restarting the server—involved in part the ad-hoc response header interposition position point I was using.

I'll try to make time to look at this more closely.

Bogdanp:
I wonder if it wouldn't be better to extend `connection` with an additional `response-processor-proc` field instead of using a param. Your new dispatcher could then add its logger by augmenting (composing its processor onto the processor field of a `struct-copy`ed struct) the connection before passing it down to the next dispatcher. That way, `output-response` could grab the processor directly off of the `conn` arg, and we wouldn't need to worry about parameterizations going wrong.

Re. performance, I would suggest setting up a minimal server that displays some text and throwing some load at it with [vegeta](https://github.com/tsenart/vegeta)[1]. Raise your file descriptor limits before running the server and vegeta (`ulimit -n unlimited` in each terminal session), then progressively hit it as hard as you can on the master branch until you find its limit, then compare that against your branch with or without the new dispatcher. My guess is the change to `output-response` won't make much of a difference (though using a field instead of a param will probably be faster). 

[1]: I like to use `tee` in the middle of my vegeta pipelines so I can easily save reports and see the results with the same command: `cat targets | vegeta attack -insecure -duration=30s -rate 100/s | tee report | vegeta report`

web-server-lib/web-server/http/response.rkt
```diff
@@ -103,7 +106,10 @@
      #'(let ([out-id out-e])
          write-e ...)]))
 
-(define (output-response-head conn bresp [chunked? #f])
+(define (output-response-head conn raw-bresp [chunked? #f])
+  (define bresp ((current-header-handler)
+                 (struct-copy response raw-bresp
+                              [output void])))
```
Bogdanp, line 112: I think voiding the `output` here makes it so no response can output any data (see `output-response-body` below). It should be safe to pass the unmodified response to the handler proc, and it should be the handler proc's responsibility not to do anything silly with it.

Bogdanp: Oh, ignore this. I misread the surrounding code, not realizing this was only outputting the headers. Still, it might make sense to allow processors to alter the output, too.

sorawee: This `bresp` is used only inside this function. `response-output` in `output-response-body` will be the original one. It is unaffected.

sorawee:
I thought about modifying `connection` before, but judging from how it's used and how stateful it is, I don't think that it is the right place. 

The top-level of the web-server is this:

```
(define conn
    (new-connection cm ip op (current-custodian) #f))

  (with-handlers ([exn-expected?
                   (lambda (_)
                     (kill-connection! conn))])
    (let connection-loop ()
      (define-values (req close?)
        (config:read-request conn config:port port-addresses))
      (reset-connection-timeout! conn config:response-timeout)
      (set-connection-close?! conn close?)
      (config:dispatch conn req)
      (if (connection-close? conn)
          (kill-connection! conn)
          (connection-loop))))
```

A connection is established once, and then it is used in a loop which could potentially invokes `dispatch` multiple times. Hence, a naive dispatcher that installs logging would install it multiple times. We of course can avoid that, but this perhaps suggests that `conn` is not the right place to store the information.

web-server-lib/web-server/http/response.rkt
sorawee: I replied too late :)

The reason why I only target `header` as opposed to the whole `response` is that `output-file` (which in turn calls `output-file/boundary`) creates a fake response with correct header but empty body, and then uses `output-response-head` to print the header directly. Giving an appearance that the `output` field could be modified would be misleading for such cases.

web-server-lib/web-server/dispatchers/dispatch-logresp.rkt
```diff
@@ -0,0 +1,127 @@
...
+            (time ,(current-seconds))
+            (code ,(response-code resp)))))
+
+(define (make-log-message log-path-or-port format-reqresp)
+  (define log-ch (make-async-channel))
+  (define log-thread
+    (thread/suspend-to-kill
+     (lambda ()
+       (let loop ([log-p #f])
+         (sync
```
Bogdanp, line 95: If you're not planning to handle other events here, you could drop the `sync`+`handle-evt` bits and just call `async-channel-get`. Additionally, you could use the thread's built-in mailbox (`thread-send`, `thread-receive`) instead of an async channel.

Also, note that using a buffered channel (whether it's an async channel or the thread's mailbox) here means that pending logs can pile up if disk IO is slow. If the request rate is greater than the rate at which things can be logged, these queues will grow without bound (Little's law). I would use a regular channel instead to create back pressure and to make the cost of using the dispatcher clear.

sorawee: FYI: this code was copied from https://github.com/racket/web-server/blob/master/web-server-lib/web-server/dispatchers/dispatch-log.rkt#L84. If I have to write it from scratch, I would not have written it this way either. 

For now, I think I will avoid touching this bit of code, since I'm not confident if there is any subtle issue that the code is handling. However, feel free to submit a PR to improve it afterwards. Alternatively, you can submit a PR to improve `dispatch-log.rkt` now -- and I will reflect the change to `dispatch-logresp.rkt`.

sorawee:
@LiberalArtist I tried your minimal example program, and it appears that `current-header-handler` is not affected by the issue:

```
;; common.rkt

    #lang racket
    (provide my-param)
    (define my-param
      (make-parameter #f))

;; servlet.rkt

    #lang web-server
    (require "common.rkt")
    (provide start)
    (define (start req)
      (response/xexpr
       `(html (head (title "Example"))
              (body (p ,(format "~v" (my-param)))))))

;; run.rkt

#lang racket
(require "common.rkt"
         "servlet.rkt"
         web-server/http
         web-server/servlet-dispatch
         web-server/http/response)

(define ((make-header-handler orig request) resp)
  (orig (struct-copy response resp
                     [message (string->bytes/utf-8 (~s 'hello:
                                                       (request-uri request)))])))

(define ((wrap-dispatcher inner) connection request)
  (define orig (current-header-handler))
  (parameterize ([my-param (request-uri request)]
                 [current-header-handler (make-header-handler orig request)])
    (inner connection request)))

(serve/launch/wait (λ (quit-sema)
                     (wrap-dispatcher
                      (dispatch/servlet start
                                        #:stateless? #t)))
                   #:banner? #10 
```

And here are the results

```
$ curl http://localhost:8000/foo --verbose
*   Trying 127.0.0.1:8000...
* Connected to localhost (127.0.0.1) port 8000 (#0)
> GET /foo HTTP/1.1
> Host: localhost:8000
> User-Agent: curl/7.79.1
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 hello: #(struct:url #f #f #f #f #t (#(struct:path/param "foo" ())) () #f)
< Transfer-Encoding: chunked
< Server: Racket
< Last-Modified: Sun, 12 Jun 2022 21:00:08 GMT
< Date: Sun, 12 Jun 2022 21:00:08 GMT
< Content-Type: text/html; charset=utf-8
<
* Connection #0 to host localhost left intact
<html><head><title>Example</title></head><body><p>(url #f #f #f #f #t (list (path/param "foo" '())) '() #f)</p></body></html>⏎

$ curl http://localhost:8000/bar --verbose
*   Trying 127.0.0.1:8000...
* Connected to localhost (127.0.0.1) port 8000 (#0)
> GET /bar HTTP/1.1
> Host: localhost:8000
> User-Agent: curl/7.79.1
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 hello: #(struct:url #f #f #f #f #t (#(struct:path/param "bar" ())) () #f)
< Transfer-Encoding: chunked
< Server: Racket
< Last-Modified: Sun, 12 Jun 2022 21:00:12 GMT
< Date: Sun, 12 Jun 2022 21:00:12 GMT
< Content-Type: text/html; charset=utf-8
<
* Connection #0 to host localhost left intact
<html><head><title>Example</title></head><body><p>(url #f #f #f #f #t (list (path/param "foo" '())) '() #f)</p></body></html>⏎
```

Note that both "struct:path/param"s in the response headers are correct. The bodies are of course incorrect, as you indicated in the mailing list.

web-server-lib/web-server/dispatchers/dispatch-logresp.rkt
jeapostrophe: Can you NOT the copy the code, but abstract it?

sorawee: @jeapostrophe I will wait until #120 is merged first, since this PR and that PR both touch the same file, and I don't want to create a conflict.

sorawee:
This is now ready for review.

web-server-lib/web-server/dispatchers/dispatch-logresp.rkt
sorawee: Done.

jeapostrophe (approved):

jeapostrophe:
Thank you!

PR #120, opened 2022-06-13
dispatch-log: use unbuffered channels, detect rotation via fs events

Bogdanp:
Prompted by the discussion in #119. This is only a slight improvement, so feel free to ignore if you don't think the change is worth it. I benchmarked the change by hitting [this minimal app](https://github.com/Bogdanp/racket-bench-ws-logging/blob/10b92df7cb20b24ff641d599d6db8fef121f9bb6/app.rkt) running on a basic Digital Ocean droplet at 1k RPS for two minutes against Racket 8.5.  The logs were written to an attached block storage volume and the attacking script rotated the log file once every 30 seconds.  Aside from the small performance improvement, the change to use an unbuffered channel should make logging a little safer under load since logs can't pile up in the queue if storage is slow (though I don't have sufficiently slow storage to test this with, so this point might be moot).

Master branch without logging:

```
Requests      [total, rate, throughput]         120000, 1000.01, 1000.00
Duration      [total, attack, wait]             2m0s, 2m0s, 692.047µs
Latencies     [min, mean, 50, 90, 95, 99, max]  374.864µs, 828.88µs, 580.348µs, 779.622µs, 1.331ms, 7.143ms, 65.443ms
Bytes In      [total, mean]                     600000, 5.00
Bytes Out     [total, mean]                     0, 0.00
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:120000
Error Set:
```

Master branch with `dispatch-log`:

```
Requests      [total, rate, throughput]         120000, 1000.02, 1000.00
Duration      [total, attack, wait]             2m0s, 2m0s, 1.285ms
Latencies     [min, mean, 50, 90, 95, 99, max]  355.839µs, 7.235ms, 638.343µs, 6.947ms, 16.195ms, 265.735ms, 572.065ms
Bytes In      [total, mean]                     600000, 5.00
Bytes Out     [total, mean]                     0, 0.00
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:120000
Error Set:
```

This branch with `dispatch-log`:

```
Requests      [total, rate, throughput]         120000, 1000.03, 1000.02
Duration      [total, attack, wait]             2m0s, 2m0s, 622.42µs
Latencies     [min, mean, 50, 90, 95, 99, max]  343.949µs, 982.759µs, 598.051µs, 1.088ms, 1.988ms, 6.797ms, 182.496ms
Bytes In      [total, mean]                     600000, 5.00
Bytes Out     [total, mean]                     0, 0.00
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:120000
Error Set:
```

Memory use is about the same between the two versions:

Screenshot 2022-06-13 at 10 40 56

(the y axis is milliseconds since program start, the "updated" line is cut short because there were no more GC logs beyond that point)

### DIFF
--- a/web-server-lib/web-server/dispatchers/dispatch-log.rkt
+++ b/web-server-lib/web-server/dispatchers/dispatch-log.rkt
@@ -1,12 +1,13 @@
 #lang racket/base
+
 (require net/url
-         (prefix-in srfi-date: srfi/19)
+         racket/contract
          racket/date
-         racket/async-channel
-         racket/match
-         racket/contract)
-(require web-server/dispatchers/dispatch
+         racket/path
+         (prefix-in srfi-date: srfi/19)
+         web-server/dispatchers/dispatch
          web-server/http)
+
 (define format-req/c (request? . -> . string?))
 (define log-format/c (symbols 'parenthesized-default 'extended 'apache-default))
 
@@ -31,7 +32,7 @@
         (log-format->format format)
         format))
   (define log-message (make-log-message log-path final-format))
-  (lambda (conn req)
+  (lambda (_conn req)
     (log-message req)
     (next-dispatcher)))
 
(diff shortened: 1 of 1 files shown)


samth:
The improvement to the distribution of response times seems quite significant. 6x better at p90!

PR #121, opened 2022-06-15
more optimizations

Bogdanp:
I'm opening this as a draft in case folks want to weigh in on the changes. I've still got a few changes I want to make over the next couple weeks so this isn't quite ready yet. This PR includes the commit from #120, but that can be merged independently so it doesn't block #119.

The PR may be easier to read if you read the commits independently. I forgot @samth also worked on removing `run-server` until I looked at the list of open PRs just now, but I took that a little further by also improving the server loop in a couple ways so I think this PR supersedes #106.

Tested against the `/plaintext` endpoint of [this app](https://github.com/Bogdanp/FrameworkBenchmarks/blob/87870c5c1d746e8c253fdd3bf03e18eec4dafc69/frameworks/Racket/racket/app.rkt). Master branch:

```
Requests      [total, rate, throughput]         30000, 1000.04, 1000.03
Duration      [total, attack, wait]             29.999s, 29.999s, 286.417µs
Latencies     [min, mean, 50, 90, 95, 99, max]  229.205µs, 318.265µs, 293.078µs, 367.174µs, 393.675µs, 606.119µs, 14.015ms
Bytes In      [total, mean]                     390000, 13.00
Bytes Out     [total, mean]                     0, 0.00
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:30000  
Error Set:
```

(note that this app uses its own [dispatcher](https://github.com/Bogdanp/FrameworkBenchmarks/blob/87870c5c1d746e8c253fdd3bf03e18eec4dafc69/frameworks/Racket/racket/app.rkt#L217-L218), not `dispatch/servlet`, which is why these numbers are better than the baseline numbers in commit 4506e7290b30d4387f6c63a0f8c95b97782fecab -- `dispatch/servlet` needs to do a lot of work that applications that don't use continuations don't benefit from)

This branch:

```
Requests      [total, rate, throughput]         30000, 1000.04, 1000.03
Duration      [total, attack, wait]             29.999s, 29.999s, 262.316µs
Latencies     [min, mean, 50, 90, 95, 99, max]  197.619µs, 253.639µs, 248.664µs, 280.93µs, 297.689µs, 344.797µs, 929.773µs
Bytes In      [total, mean]                     390000, 13.00
Bytes Out     [total, mean]                     0, 0.00
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:30000
Error Set:
```

For comparison, here's how Node 17 running express does on the same benchmark on my machine:

```
Requests      [total, rate, throughput]         30000, 1000.03, 1000.03
Duration      [total, attack, wait]             29.999s, 29.999s, 182.925µs
Latencies     [min, mean, 50, 90, 95, 99, max]  136.996µs, 190.037µs, 185.03µs, 208.014µs, 219.565µs, 303.124µs, 1.406ms
Bytes In      [total, mean]                     390000, 13.00
Bytes Out     [total, mean]                     0, 0.00
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:30000
Error Set:
```

### DIFF
--- a/web-server-lib/web-server/http/request-structs.rkt
+++ b/web-server-lib/web-server/http/request-structs.rkt
@@ -1,24 +1,22 @@
 #lang racket/base
 
-(require racket/contract
-         racket/serialize
+(require (for-syntax racket/base
+                     racket/struct-info
+                     syntax/transformer)
+         net/url-structs
+         racket/contract
          racket/match
          racket/port
          racket/promise
-         net/url-structs
-         web-server/private/util
-         (for-syntax racket/base
-                     racket/struct-info
-                     racket/match
-                     syntax/transformer))
-
-;; Most serializable structs are protected
-;; with struct-guard/c rather than contract-out.
+         racket/serialize
+         web-server/private/util)
 
 (provide
  ;; header
- (struct-out header)
  (contract-out
+  [struct header
+    ([field bytes?]
+     [value bytes?])]
   [headers-assq
    (-> bytes? (listof header?) (or/c #f header?))]
   [headers-assq*
@@ -38,10 +36,13 @@
    (-> request? (listof binding?))])
  ;; ----------------------------------------
  ;; binding
- (struct-out binding)
- (struct-out binding:form)
  binding:file/port?
  (contract-out
+  [struct binding
+    ([id bytes?])]
+  [struct (binding:form binding)
+    ([id bytes?]
+     [value bytes?])]
   [struct (binding:file binding)
     ;; enforce the `binding:file-content` contract to be extra careful
     ([id bytes?]
@@ -57,21 +58,30 @@
   [bindings-assq-all
    (-> bytes? (listof binding?) (listof binding?))]))
 
-;                                        
-;   ;;                      ;;           
-;   ;;                      ;;           
+(module+ private
+  (provide
+   (struct-out request)
+   (struct-out header)
+   headers-assq
+   headers-assq*
+   binding?
+   (struct-out binding:form)
+   make-binding:file/port))
+
+;
+;   ;;                      ;;
+;   ;;                      ;;
 ;   ;;;;;   ;;    ;;     ;;;;;  ;;   ;; ;
-;   ;;  ;  ;  ;  ;  ;   ;   ;; ;  ;  ;;; 
-;   ;;  ;; ;  ;     ;;  ;   ;; ;  ;  ;;  
-;   ;;  ;;;;;;;;  ;;;; ;;   ;;;;;;;; ;;  
-;   ;;  ;; ;     ;  ;;  ;   ;; ;     ;;  
-;   ;;  ;; ;    ;;  ;;  ;   ;; ;     ;;  
-;   ;;  ;;  ;;;  ;;; ;   ;;; ;  ;;;  ;;  
-;                                        
+;   ;;  ;  ;  ;  ;  ;   ;   ;; ;  ;  ;;;
+;   ;;  ;; ;  ;     ;;  ;   ;; ;  ;  ;;
+;   ;;  ;;;;;;;;  ;;;; ;;   ;;;;;;;; ;;
+;   ;;  ;; ;     ;  ;;  ;   ;; ;     ;;
+;   ;;  ;; ;    ;;  ;;  ;   ;; ;     ;;
+;   ;;  ;;  ;;;  ;;; ;   ;;; ;  ;;;  ;;
+;
 
 (define-serializable-struct header
   (field value)
-  #:guard (struct-guard/c bytes? bytes?)
   #:transparent)
 
 (define (headers-assq* f hs)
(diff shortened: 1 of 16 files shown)


web-server-lib/web-server/safety-limits.rkt
```diff
@@ -65,6 +69,7 @@
 
 
 (define-safety-limits/private-submodule
+  max-concurrent positive-count/c 1000 #:unlimited +inf.0
```
Bogdanp, line 72: This limit may be too conservative, so it might be worth raising.  It's a hard balance, though:

* if there's no limit, then it's easy to DoS the server, 
* if the default limit is too low, we may be limiting performance on very fast servers
* if the default limit is too high, slow servers may be open to DoS

IME, once you have several thousands of threads running, performance tends to degrade so `1000` feels right. Arguably, people doing very high performance stuff w/ the web server should be aware of and tweaking `safety-limits` anyway. I'll have to take a look at what other servers' defaults are.

cc @LiberalArtist

samth:
Does removing `run-server` mean it no longer depends on compatibility-lib?

jeapostrophe (commented):
This looks awesome to me

Bogdanp:
> Does removing run-server mean it no longer depends on compatibility-lib?

Yup, looks like it does.

LiberalArtist:
Re replacing `run-server`:

It's a pathological case, but mechanisms like `(kill-thread (current-thread))` and `(custodian-shutdown-all (current-custodian))` escape without running winders, e.g.:
```racket
#lang racket
(dynamic-wind
 void
 (λ ()
   (kill-thread (current-thread)))
 (λ ()
   (println 'this-never-prints)))
```
I think `run-server`'s cleanup thread handles this by `sync`ing on the handler thread (without the vestigial timeout, it might be more efficient to go via `thread-dead-evt`), but I think this replacement would not call `custodian-shutdown-all` on the custodian that owns the ports.

This sort of subtlety is why I suggested in <https://github.com/racket/web-server/pull/106#issuecomment-882132036>:
> If it seems better to copy–paste for now, I'd add a comment here and maybe even in `mzlib/thread` in the hope of prompting someone to tackle the refactoring in the future.

More broadly, this is reminding me of mailing-list discussion in https://groups.google.com/g/racket-users/c/AAeEp_llxaY/m/uRviksPGAgAJ and https://groups.google.com/g/racket-dev/c/PyetqHSjtAA/m/slBdazdqAwAJ about the subtleties of delimited control and the Concurrent ML subsystem and the limitations of finalization with `dynamic-wind`. (I'd recently recommended these threads to people adapting the [Shepherd](https://www.gnu.org/software/shepherd/manual/shepherd.html) (a daemon-managing daemon) to use Guile's Fibers.) In particular, @lexi-lambda pointed out the technical report [“Algebraic Effect
Handlers with Resources and Deep Finalization”](https://www.microsoft.com/en-us/research/uploads/prod/2018/04/resource-v1.pdf) (which I want to finish reading!), which says in §8.7 "Dynamic Wind" (internal citations omitted):

> The Scheme language always supported delimited continuations and
> they have also struggled with initialization- and finalization
> for continuations that resumed more than once. The
> `unwind-protect` in Scheme is like a `finally` clause, while
> `dynamic-wind` is like `initially`/finally with a pre- and
> postlude. [Sitaram](https://web.archive.org/web/20200223020813/http://www.ccs.neu.edu/~dorai/uwcallcc/uwcallcc.html) describes how the standard dynamic-wind is not
> good enough in general:
>
>> While this may seem like a natural extension of the first-order
>> `unwind-protect` to a higher-order control scenario, it does not
>> tackle the pragmatic need that `unwind-protect` addresses,
>> namely, the need to ensure that a kind of “clean-up” happens only
>> for those jumps that significantly exit the block, and not for
>> those that are a minor excursion.  The crux is identifying which
>> of these two categories a jump falls into.
>
> Interestingly, this is exactly what is addressed by algebraic
> effects where “significant exit”s are operations that do not
> resume, while “minor excursions” are regular operations that
> resume with a result.

I'll try to take a look at other aspects of this PR later.

Bogdanp:
The `(custodian-shutdown-all (current-custodian))` case is covered by the custodian that wraps the handler, but you're right about the `(kill-thread (current-thread))` case. I don't think the server claims that handler threads are kill-safe, but it's easy enough to bring back the cleanup thread so I'll just do that. An alternative I've considered is keeping track of the handler threads in a list inside the loop and syncing on them to remove them from the set when done. That would get rid of the need for both `dynamic-wind` and a cleanup thread, but would result in more work per loop iteration so it's probably not worth it.

> I think this replacement would not call custodian-shutdown-all on the custodian that owns the ports.

I'm not sure I understand what you mean here. The custodian that gets shut down is the one that wraps both the thread and the TCP ports.

web-server-lib/web-server/http/request-structs.rkt
```diff
@@ -57,21 +58,30 @@
...
+   (struct-out header)
+   headers-assq
+   headers-assq*
+   binding?
+   (struct-out binding:form)
+   make-binding:file/port))
+
+;
+;   ;;                      ;;
+;   ;;                      ;;
```
Bogdanp, line 73: I've included these whitespace changes because it was pain to work around them, but I can redo the commit if anyone cares. Actually, what I'd really like to do is remove these comments because I think they make reading the code harder and I can't really even read the ascii art (and I assume I'm not alone). Thoughts?

web-server-lib/web-server/http/request-structs.rkt
```diff
@@ -84,49 +94,47 @@
...
-;                                    ;;   ; 
-;                                      ;;;  
-;                                           
+;                                    ;;   ;
+;                                      ;;;
+;
 
 (define-serializable-struct binding
   (id)
-  #:guard (struct-guard/c bytes?)
```
Bogdanp, line 129: Dropping these guards led to (by far) the largest improvement in the whole PR. This is slightly less safe for data that's deserialized, but I think safety is preserved outside this module because I've added `contract-out` forms for these structs (so, even if someone were to deserialize headers with invalid data, this module would be blamed on access to that data).

Bogdanp:
Here's how we're doing so far:

* [master branch](https://www.techempower.com/benchmarks/#section=test&shareid=85490882-b757-4fa0-9340-5babbb12ffc6&hw=ph&test=composite&a=2) (88,090RPS on the plaintext benchmark)
* [latest commit on this branch](https://www.techempower.com/benchmarks/#section=test&shareid=67e37271-e872-44a8-96f1-557e38b7fa06&hw=ph&test=json&a=2) (151,709RPS on the plaintext benchmark)

For comparison, here's the plaintext benchmark on [express.js](https://www.techempower.com/benchmarks/#section=test&shareid=ec422152-d3bc-46b0-9a04-00b46a871f34&hw=ph&test=plaintext&a=2) on the same machine. The test machine in this case is a 24-core AMD server.

web-server-lib/web-server/http/request-structs.rkt
samth: what happens if the guard is `(lambda (x y) (if (bytes? x) x (error 'fail)))`?

Bogdanp: It's [not bad](https://www.techempower.com/benchmarks/#section=test&shareid=8d0719f4-2571-4931-bf9a-05e6697e95bc&hw=ph&test=composite&a=2) ([diff](https://github.com/racket/web-server/commit/a49d469aaa6cca530daddda19b0efd8b2c4ccde4)), but it doesn't feel great to lose 4k RPS to this. 

My understanding is these structs are serializable so that they can be used in serialized continuations via stuffers and anyone serializing continuations to URLs should be using the HMAC stuffer to prevent tampering (otherwise they probably have much bigger problems), so I don't know if these guards actually buy us much safety. That said, I don't feel too strongly about this and between this and the custom deserializer route, I definitely think this is preferable.

web-server-lib/web-server/http/request-structs.rkt
jeapostrophe: Go for it :)

web-server-lib/web-server/http/request-structs.rkt
jeapostrophe: I don't think it is worth it

web-server-lib/web-server/http/request.rkt
```diff
@@ -140,8 +135,23 @@
...
+         [(16)
+          (for/fold ([n 0])
+                    ([b (in-bytes bs)] #:when n)
+            (or
+             (and (>= b 48) (<= b 57)  (+ (* n 16) (- b 48))) ;; 0-9
+             (and (>= b 65) (<= b 70)  (+ (* n 16) (- b 55))) ;; A-F
+             (and (>= b 97) (<= b 102) (+ (* n 16) (- b 87))) ;; a-f
+             ))]
+         [else
+          (raise-argument-error 'bytes->number "(or/c 10 16)" base)])))
```
jeapostrophe, line 154: This change seems like something that should be integrated into Racket core

jeapostrophe: (or at least a utility library)

samth: Seems similar to integer-bytes->integer.

Bogdanp: The closest thing in core to this function is `string->number`. `integer-bytes->integer` is meant for interpreting a series of bytes as an integer, so it wouldn't be appropriate here:

```racket
> (bytes->nonnegative-integer #"123")
123
> (integer-bytes->integer #"123" #t)
integer-bytes->integer: length is not 1, 2, 4, or 8 bytes
  length: 3
 [,bt for context]
> (integer-bytes->integer #"1234" #t)
875770417
> (integer-bytes->integer #"\xFF\x00\x00\x00" #t #f)
255
```

I agree that it would be nice to have a `bytes->number` in core, though. I'll see if I can find time to do that after this PR. This specific version, though, is specialized to what the web server needs (doesn't support negative numbers, floats), so it's more work than just copying it over.

jeapostrophe: Ok

jeapostrophe:
I think we should merge and brag ;)

Bogdanp:
I'll wrap this up tomorrow morning and then we can merge. There are more improvements that could be made, but we can leave those for another time so this doesn't get too large and because I'm not sure when I'll have time to do them.

web-server-lib/web-server/http/request-structs.rkt
Bogdanp: I'll do this next time I work on this module.

web-server-lib/web-server/safety-limits.rkt
Bogdanp: In the end, I bumped this to 10k. Most of the servers I looked at had no default limit (but they did have ways to enforce a limit), except for gunicorn which has a limit of 1k by default.

Bogdanp:
OK, I think this is good to go now.

PR #128, opened 2023-09-01
Add Combined Log Format option

camilledalmeras:
The Combined Log Format adds "Referer" and "User-Agent" fields to the Apache default log format.

Compliance with the Combined Log Format definition is not 100%, as the size of the object returned to the client is missing, but this caveat is duly pointed out in the added docs.

Thanks for your work on Racket’s web server!


jeapostrophe (approved):

jeapostrophe:
Thank you!!

PR #133, opened 2024-03-14
Update continue.scrbl

timbot1789:
A key of size `1024` is no longer considered long enough to be secure. Running the line `plt-web-server --ssl` throws the following error on recent versions:

`ssl-load-certificate-chain!: load failed from: #<path:/myPath/Racket/server-cert.pem> (error:140AB18F:SSL routines:SSL_CTX_use_certificate:ee key too small)`

Bumping the size up to `2048` fixes the error.

---
Kudos on the excellent tutorial. I walked through every line, and this was the only issue I ran into.


jeapostrophe (approved):

jeapostrophe:
Thank you!

PR #134, opened 2024-04-29
docs: link to wrapped make-cookie

benknoble:
This greatly aids traversing the documentation: readers of this module's make-cookie can now jump to the documentation of the wrapped function instead of jumping to net/cookies/server and scrolling or searching.

The only downside is that the link will display as "net:make-cookie" instead of "make-cookie"; see the GUI Easy documentation [1] as an example (e.g., "gui:dc<%>" in "Custom Views"). In that document, all racket/gui exports are prefixed with gui: to avoid ambiguity, and the links take you to the correct place within racket/gui documentation. That is accomplished by

    (require (for-label (prefix-in gui: racket/gui)))

which I wanted to avoid here: I didn't want all references to net/cookies/server to use the net: prefix.

[1]: https://docs.racket-lang.org/gui-easy/index.html

### DIFF
--- a/web-server-doc/web-server/scribblings/http.scrbl
+++ b/web-server-doc/web-server/scribblings/http.scrbl
@@ -454,6 +454,13 @@ transmission that the server @bold{will not catch}.}
                      web-server/http/response-structs
                      web-server/http/cookie))
 
+@(begin
+   (define-syntax-rule (define-from-net/cookies/server name export)
+     (begin
+       (require (for-label net/cookies/server))
+       (define name @racket[export])))
+   (define-from-net/cookies/server net:make-cookie make-cookie))
+
 @(define rfc6265
    (hyperlink "https://tools.ietf.org/html/rfc6265.html"
               "RFC 6265"))
@@ -474,7 +481,7 @@ transmission that the server @bold{will not catch}.}
           cookie?]{
    Constructs a cookie with the appropriate fields.
 
-   This is a wrapper around @racket[make-cookie] from @racketmodname[net/cookies/server]
+   This is a wrapper around @|net:make-cookie| from @racketmodname[net/cookies/server]
    for backwards compatibility. The @racket[comment] argument is ignored.
    If @racket[expires] is given as a string, it should match
    @link["https://tools.ietf.org/html/rfc7231#section-7.1.1.2"]{RFC 7231, Section 7.1.1.2},


sorawee:
@mflatt showed this trick to me earlier: you can do

https://github.com/racket/racket/blob/master/pkgs/racket-doc/syntax/scribblings/module-reader.scrbl#L10

to locally `@racket[make-cookie]` where `make-cookie` is from `net/cookies/server`.

benknoble:
Unclear what maintainers to ping for review/merge, so I'll start with @jeapostrophe given his name in `pkg-authors`.

jeapostrophe:
I think it is important for the link to be correct, thank you for the patch. But, I think it will be ugly for it to say `net:make-cookie`, so I prefer Sorawee's trick. I think it is worth putting in the macro without the abstraction if you're willing @benknoble

benknoble:
Adjusted. range-diff:

```
1:  2594314 ! 1:  4bdb749 docs: link to wrapped make-cookie
    @@ Commit message
         make-cookie can now jump to the documentation of the wrapped function
         instead of jumping to net/cookies/server and scrolling or searching.
     
    -    One possible solution is to prefix the special name like
    -    "net:make-cookie": the only downside is that the link will display as
    -    "net:make-cookie" instead of "make-cookie"; see the GUI Easy
    -    documentation [1] as an example (e.g., "gui:dc<%>" in "Custom Views").
    -    In that document, all racket/gui exports are prefixed with gui: to avoid
    -    ambiguity, and the links take you to the correct place within racket/gui
    -    documentation. That is accomplished by
    +    The only downside is that the link will display as "net:make-cookie"
    +    instead of "make-cookie"; see the GUI Easy documentation [1] as an
    +    example (e.g., "gui:dc<%>" in "Custom Views"). In that document, all
    +    racket/gui exports are prefixed with gui: to avoid ambiguity, and the
    +    links take you to the correct place within racket/gui documentation.
    +    That is accomplished by
           (require (for-label (prefix-in gui: racket/gui)))
         which I wanted to avoid here: I didn't want all references to
    -    net/cookies/server to use the net: prefix, and it is desirable for the
    -    docs to show "make-cookie".
    -
    -    Thus instead we use a trick (by way of Sorawee and Matthew Flatt) to
    -    embed the require-for-label in a macro use [2].
    +    net/cookies/server to use the net: prefix.
     
         [1]: https://docs.racket-lang.org/gui-easy/index.html
    -    [2]: https://github.com/racket/racket/blob/f2294a69784a5e77e83d8cf07d6cddab7f1b09fd/pkgs/racket-doc/syntax/scribblings/module-reader.scrbl#L10
     
      ## web-server-doc/web-server/scribblings/http.scrbl ##
     @@ web-server-doc/web-server/scribblings/http.scrbl: transmission that the server @bold{will not catch}.}
    -                      web-server/http/response-structs
    -                      web-server/http/cookie))
    + @; ------------------------------------------------------------
    + @section[#:tag "cookie"]{Placing Cookies}
      
    -+@(begin
    -+   (define-syntax-rule (define-from-net/cookies/server name export)
    -+     (begin
    -+       (require (for-label net/cookies/server))
    -+       (define name @racket[export])))
    -+   (define-from-net/cookies/server net:make-cookie make-cookie))
    -+
    - @(define rfc6265
    -    (hyperlink "https://tools.ietf.org/html/rfc6265.html"
    -               "RFC 6265"))
    +-@(require (for-label (except-in net/cookies/server
    +-                                make-cookie)
    ++@(require (for-label (rename-in net/cookies/server
    ++                                [make-cookie net:make-cookie])
    +                      net/cookies/common
    +                      web-server/servlet
    +                      web-server/http/xexpr
     @@ web-server-doc/web-server/scribblings/http.scrbl: transmission that the server @bold{will not catch}.}
                cookie?]{
         Constructs a cookie with the appropriate fields.
      
     -   This is a wrapper around @racket[make-cookie] from @racketmodname[net/cookies/server]
    -+   This is a wrapper around @|net:make-cookie| from @racketmodname[net/cookies/server]
    ++   This is a wrapper around @racket[net:make-cookie] from @racketmodname[net/cookies/server]
         for backwards compatibility. The @racket[comment] argument is ignored.
         If @racket[expires] is given as a string, it should match
         @link["https://tools.ietf.org/html/rfc7231#section-7.1.1.2"]{RFC 7231, Section 7.1.1.2},
```

jeapostrophe (approved):

jeapostrophe:
Thank you!

benknoble:
Unrelated, but it looks like CI is failing because a test takes too long?

sorawee:
Yeah, the CI has been failing for a while now. It's not due to this PR.

PR #138, opened 2025-02-08
Change private key loading to default to PEM-encoded PKCS#8

bdj:
Simpler change to support LetsEncrypt keys in the webserver. See https://github.com/racket/web-server/pull/137 for more discussion. If someone has better copy for the documentation, I would be happy to change it. I have tested this for my use case and it works, but I would also be willing to add tests if there is a good way to test this.

cc @LiberalArtist @rmculpepper


bdj:
Is there anything else that needs to be done here to merge this?

PR #139, opened 2025-07-19
server: add support for graceful shutdown

Bogdanp:
This change adds a new safety limit, `#:shutdown-grace-period`, which controls how long the server will wait for in-progress connections to drain before shutting down. The current behaviour of shutting down immediately is preserved by default, so this change is opt-in.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,7 @@ jobs:
     name: "Build on Racket '${{ matrix.racket-version }}' (${{ matrix.racket-variant }})"
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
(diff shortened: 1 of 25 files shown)


Bogdanp:
I think the "gh3" failure is due to an incompatibility between old versions of Racket and new versions of OpenSSL. After some experimenting, I got the following error: https://github.com/Bogdanp/web-server/actions/runs/16388725990/job/46311858304#step:6:97

I'll push a change to add some more recent versions of Racket to the test matrix and to disable that test for older versions.

Bogdanp:
👋 bump

jeapostrophe (approved):

jeapostrophe:
Thank you!

